PR #3, opened 2015-12-22
Update gluestick to have a proper production mode

toddw:
Production mode bundles the assets so you can load them from a CDN, it disables hot module replacement and doesn't use webpack dev tools for static file serving.


christinebrass:
:+1:

truecarhelpdesk:
Please take us off this thread
[cid:50647101-C477-49CB-AEB7-73BF8C4A831B]

PR #6, opened 2015-12-28
Refine watch to avoid unnecessary server restarts.

threehams:
Only restart the server when src/ and assets/ directories change. This avoid constant server restarts when using WebStorm or any other JetBrains IDE (the existing watch included dot files / dot directories recursively).

Also added .idea/ to .gitignore (JetBrains IDE files).


toddw:
Thanks @threehams Can you include the 'Index.js' file that lives at the root? I'm thinking about moving that file but I don't want to hold up this change while we decide if/where to put it. I think you could just watch that one file manually for now.

threehams:
After testing this a bit more, I'm noticing that watches are very unreliable in OS X. I've replaced the fs.watch with chokidar - this is the same package used by webpack for file watching, so it's not really adding any extra dependencies.

After this change, I'm able to drop the timeout down to 250ms with no issues. It may be possible to go down to 100ms or lower, but it'd be nice to try other environments first.

toddw:
This is great! Let me test this out and make sure there aren't any side effects but I doubt there will be. I'll have this merged and published on a new point version by the end of the day.

toddw:
Sorry for the delay @threehams, we are putting together a contributor agreement much like Facebook has done with React (and all their other open source projects). You were the first pull request so the issue hadn't come up yet. Once we get that finished things should move along much quicker.

I really do appreciate this PR and I can't wait to get it merged.

toddw:
Please resolve merge conflicts…

PR #9, opened 2016-01-06
add matchMedia polyfill (IE9)

moarwick:
Add polyfills to support Radium media queries on IE9
https://github.com/paulirish/matchMedia.js


toddw:
:+1:

Issue #12, opened 2016-01-08
Warning: React attempted to reuse markup...


kitze:
I'm following the tutorial. I've added Todos container and added it to the routes in the config. 

Then i navigated to localhost:8888/todos and i got this error https://i.imgur.com/rc3m4k5.png

I've tried to remove everything related to redux from the component but the error persists.

I've tested this with node 4._, node 5._, Chrome Canary, and normal version of Chrome. Also i've tried it in incognito window without any extensions but the error is still there.

toddw:
I will work on resolving this. The resuse markup error is a bummer and I'm going to fix that but that one specifically shouldn't be causing a crash. The [Resolve] error looks more serious. I'll work on a fix for both

kitze:
@toddw the [Resolve] error is coming from the redux dev tools extension. I don't get it in incognito mode without the extension.

toddw:
Fixed with https://github.com/TrueCar/gluestick/pull/14

PR #14, opened 2016-01-08
Resolve rerender error

toddw:
There was a strange bug that was causing the server to render an extra
div that the client was not rendering. This was super weird and after
rearanging a few things I was able to resolve it.

Related to: https://github.com/TrueCar/gluestick/issues/12


christinebrass:
:+1:

PR #19, opened 2016-01-13
Render simple loading page during server restarts

christinebrass:
This PR addresses issues seen when one refreshes the page during server restarts.  

A loading screen with a spinner is now rendered instead of a page with the error "Error: connect ECONNREFUSED 127.0.0.1:8880."


toddw:
:+1:

PR #20, opened 2016-01-13
Fix shared `store` instance bug

toddw:
The way the store was being created and imported was actually creating a
single store instance that was shared on the server. This update ensures
that it generates a new store on each new web request.


christinebrass:
:+1:

PR #25, opened 2016-01-15
Add support for custom 404 handler

toddw:
New projects will be created with a custom 404 handler already included.
If this handler is removed, messed up, or if this is an older GlueStick
project, we will show some help text in the console that explains how a
custom 404 page could be created.

I would like to start consolidating more of the help text into this
file. Especially while we are moving super fast and potentially
inentially breaking things. Once we hit 1.0 we wont be able to be so
care free but let's at least try to be helpful until then.

screen shot 2016-01-14 at 6 03 46 pm


christinebrass:
:+1:

Issue #7, opened 2015-12-28
Windows support


mbinette:
Trying to use Gluestick on Windows 10 fails. It looks to be path related. There must be code in Gluestick that is using the home directory. I get errors such as:

./src/components/MasterLayout.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./C:/Users/Michael/AppData/Roaming/npm/node_modules/gluestick/node_modules/webpack/buildin/module.js in D:\dev\fas\web\src\components
resolve file
  D:\dev\fas\web\src\components\C:\Users\Michael\AppData\Roaming\npm\node_modules\gluestick\node_modules\webpack\buildin\module.js doesn't exist
  D:\dev\fas\web\src\components\C:\Users\Michael\AppData\Roaming\npm\node_modules\gluestick\node_modules\webpack\buildin\module.js.js doesn't exist

toddw:
Thanks for reporting this @mbinette! You might be the first person to try GlueStick on Windows. I'll look into it.

toddw:
@mbinette - This should be fixed now… I wasn't getting exactly the same error as you but I was getting errors. #31 resolved all of my issues.

supergibbs:
Latest Gluestick works on the Windows 10 Creator's Update using the Ubuntu subsystem. https://blogs.msdn.microsoft.com/commandline/2017/04/11/windows-10-creators-update-whats-new-in-bashwsl-windows-console/

PR #31, opened 2016-01-19
Windows support

toddw:
Although I was not able to specifically clone the exact error message from #7, I did see an array of issues when trying to run GlueStick on Windows 10. With this PR I was able to get GlueStick fully functional in Windows 10.

https://github.com/TrueCar/gluestick/issues/7


christinebrass:
:+1:

PR #32, opened 2016-01-21
Add `gsBeforeRoute` and deprecate `fetchData`

toddw:
`fetchData` is currently the only place where you can execute code
before routing to a container. Since it is not only used for fetching
data, it shouldn't be called `fetchData`. Also, some people were
confusing `fetchData` as a React method and not realizing it was
specific to GlueStick. For this reason we are going to rename the method
to `gsBeforeRoute`.

Along with the rename, we need to include arguments so that developers
will know if this method is being called from the server or from the
client.

While updating the container generator I noticed that our generator
output needed to be re-tabbed so that was done with this commit as well.

### DIFF
--- a/generate/component.js
+++ b/generate/component.js
@@ -1,12 +1,12 @@
 import React, { Component, PropTypes } from "react";
 
 export default class __$NAME__ extends Component {
-    render () {
-        return (
-            <div>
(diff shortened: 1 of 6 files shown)


christinebrass:
:+1:

PR #33, opened 2016-01-22
Improve shared package management

toddw:
1. Moved code shared by the CLI (includes server rendering engine) and the application
   code into a shared module. This makes it so projects no longer have to
   update their `gluestick` version when the CLI changes. The shared code
   changes way more infrequently and will result in less package updates
   with this strategy.
2. Moved from .x versions in package to full versions. This makes things
   more predictable.
3. Added code to the auto-upgrader that will check the project's
   package.json file and compare it to the CLI's package.json file and the
   project.json file that will be included in newly created projects. We
   then compare versions and resolve mismatches and missing modules that we
   decided new projects should have. The user will be prompted before
   updating but changes will be automatically applied and installed if they
   agree to the prompt.
4. While updating the generators I noticed that two of the containers
   were using the deprecated `fetchData` method instead of `gsBeforeRoute`.
   Updated those.
# 

screen shot 2016-01-21 at 5 55 48 pm


christinebrass:
:+1: :100:

PR #34, opened 2016-01-23
Use PM2 for server side rendering

toddw:
[PM2](http://pm2.keymetrics.io/) is a really useful process manager. It
supports cluster mode so we can run multiple instances of the app,
depending on the number of CPUs or we can manually set the number of
instances. It will automatically restart instances if they are taking up
too much memory. It gives us nice monitoring as well.


christinebrass:
:+1:

PR #35, opened 2016-01-25
fix redux dev tools

toddw:
There was no way for the dev tools to know which environment gluestick
was running in. We now expose this in a global variable and we only show
the dev tools if we are not in production mode

Related to https://github.com/TrueCar/gluestick-shared/commit/c2c655f48760e53913d9707446fb2e038dbd37d4
and
https://github.com/TrueCar/gluestick-shared/commit/2109beb26af2afd336d3c0f289599e98cc9ce4c2

### DIFF
--- a/new/package.json
+++ b/new/package.json
@@ -8,7 +8,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


christinebrass:
:+1:

PR #37, opened 2016-01-26
Add destroy to the CLI

jer-k:
I went to create a container, accidentally did `gluestick generate component something` and realized it would be nice to have a destroy method like [Rails](http://guides.rubyonrails.org/command_line.html#rails-destroy). I more or less copied what is in `src/commands/generate.js` but modified it as needed. 

Something to consider:
Steps 1 - 4 in `destroy` and `generate` are the same aside from the messages they log. Does that warrant possible extraction?

### DIFF
--- a/src/commands/generate.js
+++ b/src/commands/generate.js
@@ -108,7 +108,7 @@ module.exports = function () {
   }
 
   // Step 8: Write test file for component
-  if(["component"].indexOf(command) !== -1) {
+  if(command === "component") {
     var testPath = path.join(process.cwd(), "/test/", availableCommands[command], `/${name}.test.js`);
     var testTemplate;
     var testFileExists = true;
--- a/src/commands/generate.js
+++ b/src/commands/generate.js
@@ -87,18 +87,13 @@ module.exports = function () {
   if (command === "reducer") {
     var reducerIndexPath = path.resolve(process.cwd(), "src/reducers/index.js");
     try {
-      var indexLines = fs.readFileSync(reducerIndexPath, {encoding: "utf8"}).split("\n");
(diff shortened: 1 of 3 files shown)


src/commands/generate.js
```diff
@@ -91,7 +91,7 @@ module.exports = function () {
       var lastLineIndex = indexLines.length - 1;
       var newLine = `export { default as ${name} } from "./${name}"`;
       if (indexLines[lastLineIndex] === "") {
-        indexLines.splice(lastLineIndex - 1, 1, newLine);
+        indexLines.splice(lastLineIndex, 1, newLine);
```
jer-k: Adding a note here on how to replicate. I ran into this when testing out the destroy commands. [Array.splice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) will start deleting at the index it is supplied so by passing `lastLineIndex - 1` you could potentially move to a line with a reducer and delete it by accident.

Using gluestick before this commit

```
gluestick new testapp
gluestick generate reducer testone
```

Now `src/reducers/index.js` should look like this
screen shot 2016-01-25 at 11 36 36 pm

But sometimes I have weird tendencies and decide to delete white space and so I changed `src/reducers/index.js` to look like this:
screen shot 2016-01-25 at 11 37 24 pm

Then I ran
`gluestick generate reducer testtwo` and `src/reducers/index.js` looked like this:
screen shot 2016-01-25 at 11 38 00 pm

I'm not sure if you had specific intentions here, but I don't think having reducers overwrite each other because someone has made modifications to `src/reducers/index.js` is a desired effect.

toddw:
:+1:

PR #38, opened 2016-01-26
Added a catch all command to the CLI 

jer-k:
Sometimes we mistype commands but Gluestick was just silently doing nothing. Users should be informed that they attempted to perform an unknown command and then shown the help to provide them with known commands.
screen shot 2016-01-26 at 9 35 30 am


toddw:
:+1:

PR #39, opened 2016-01-27
Only stop processes belonging to gluestick

christinebrass:
This PR aims to prevent gluestick from stopping any non-gluestick apps that are also being managed by PM2.


toddw:
:+1:

PR #40, opened 2016-01-27
Custom redux middleware

jer-k:
User's of Gluestick may need to add custom redux middleware. This provides them with a way to do so. This also provides auto-upgrade support for this feature.

A precursor to this request is: https://github.com/TrueCar/gluestick-shared/pull/1


toddw:
:+1:

PR #44, opened 2016-01-28
Extendable Webpack

jer-k:
Some users may want to extend the Webpack settings that have been defined in Gluestick with their own Preloaders and Loaders. This change allows them to specify those Preloaders and Loaders in `webpack-additions.js` and then Gluestick will add them.

Plugins not currently supported.


toddw:
:+1:

PR #47, opened 2016-01-30
Enable css hotloading in development mode

christinebrass:
We want to use ExtractTextPlugin in prod mode, but in dev mode just use the normal style loader.


toddw:
:+1:

PR #49, opened 2016-02-01
Fix style flicker issue in dev mode

christinebrass:
Use the assets that webpack-isomorphic-tools provides to render styles in dev mode, preventing the style flash seen when reloading a page.


toddw:
:+1:

PR #50, opened 2016-02-01
Create log file if it doesn't exist

christinebrass:
Prevent PM2 log file error by opening for appending, which will create the file if it does not exist.


toddw:
:+1:

PR #51, opened 2016-02-01
Generate Reducer and Container tests

jer-k:
We should have a basic test for everything we generate to encourage users to write tests!


toddw:
:+1:

PR #52, opened 2016-02-02
Move webpack-isomorphic-tools-configuration inside src/

jer-k:
Someone reported that running the latest `develop` branch by running `npm install -g .` causes the following error.

```
/usr/local/lib/node_modules/gluestick/webpack-isomorphic-tools-configuration.js:14
const { additionalLoaders, additionalPreLoaders } = require("./src/lib/get-webpack-additions")(true);
^

SyntaxError: Unexpected token {
```

This was because webpack-isomorphic-tools-configuration.js wasn't being run through Babel so the use the const was not supported. Moving the file inside the `src` directory ensure that it is included when `run-through-babel.js` is invoked.


toddw:
:+1:

PR #53, opened 2016-02-02
Setup test script

christinebrass:
Bootstrap a test environment. We'll of course want more than one test as part of our test suite, but this just gets things setup so we can start writing more.


toddw:
:+1:

PR #54, opened 2016-02-02
Fix 404 generator and instructions

toddw:
Things got a little out of date in the generated files and missing 404
instructrions.


christinebrass:
:+1:

PR #55, opened 2016-02-04
Add more tests for `gluestick new` cli

christinebrass:
This PR simply adds a couple more tests to our CLI test suite. 

Specifics:
- use `spy` as opposed to `stub` for mocking console calls, so we still see output
- move the npm install process to a separate module so it can be more easily stubbed, in order to prevent installs during tests
- add test for case where a gluestick project already exists in the current folder
- add test that verifies the contents of the new folder was copied over correctly


toddw:
:+1:

PR #56, opened 2016-02-09
Add `dockerize` command

toddw:
Working on the `dockerize` command revealed some issues with the asset urls and bundling. This PR adds the dockerize command and fixes those issues.


new/src/config/.Dockerfile
```diff
@@ -0,0 +1,29 @@
+# DO NOT MODIFY
+# This file is automatically generated. You can copy this file and add a
+# Dockerfile to the root of the project if you would like to use a customer
+# docker setup.
+FROM node:5.5.0
+
+RUN apt-get update && \
+    apt-get -y install libjpeg62-turbo-dev libpango1.0-dev libgif-dev build-essential g++
+
+RUN npm install -g gluestick
```
toddw, line 10: A little bit of chicken/egg problem with this line. We need the version of GlueStick in this branch for this to work but we don't want to check in code that points to a branch which will be deleted. If you want to test this before it is merged replace this line with `RUN npm install TrueCar/gluestick#toddw/deployment-updates -g`

christinebrass, line 10: :+1: makes sense

christinebrass:
:+1: :100:

PR #59, opened 2016-02-11
Fix update-package

toddw:
When there is a mismatch between dependencies in the gluestick cli and
the app code, update-package prompts the user asking for permisson to
update the package file. It then runs `npm install` to update things.
This wasn't proving to be very reliable, instead we have to first wipe
the `node_modules` folder, then we run `npm install` this works better.

I was tempted to call `process.exit()` and make the user re-run
`gluestick start` but after running several tests with the wiped
`node_modules` folder, I no longer think that will be necessary.


jer-k:
:+1:

PR #60, opened 2016-02-11
Specify GlueStick version in Dockerfile

toddw:
Simply having `npm install gluestick -g` in the Dockerfile does not work
because your machine could cache the image up to that point and future
images could end up with outdated versions of GlueStick. Not only that
but it forces you to use the latest version of GlueStick on new
non-cached images when maybe you built it with an older version and
haven't upgraded yet.

This update ensures that the Docker image will install the version of
GlueStick that was used to create the image.


jer-k:
:+1:

PR #61, opened 2016-02-11
Add support to load css files from npm modules

toddw:
Use case: `import 'bootstrap/dist/css/bootstrap.css';`

This was causing problems for 2 reasons.
1. The bootstrap.css file was references fonts outside of the `assets`
   folder. Our loader was limited to only loading fonts from the `assets`
   folder. There is now a good reason why we should not limit it to the
   that folder and I couldn't think of any good reason why we should. So
   that restriction was removed from both the fonts and images. CSS files
   can reference fonts and images and we want those to work.
2. In development mode, we have a really sweet hack that gets rid of the
   flash of unstyled content. You can see it in `getHead.js`. We get a list
   of assets and do some things… but webpack-isomorphic-tools replace
   `node_modules` with `~` which was breaking that. So now it is swapped
   back to `node_modules`. This is only development mode, so don't panic if
   you don't like it but please share if there is a better way.
3. I moved the css loader stuff out of the client only side and put it in shared with all of it's friends. If there is a reason **not** to do this, now would be the time to bring it up.

### DIFF
--- a/src/commands/shared.js
+++ b/src/commands/shared.js
@@ -1,6 +1,7 @@
 var path = require("path");
 var process = require("process");
 var WebpackIsomorphicToolsPlugin = require("webpack-isomorphic-tools/plugin");
+var ExtractTextPlugin = require("extract-text-webpack-plugin");
 
 const isProduction = process.env.NODE_ENV === "production";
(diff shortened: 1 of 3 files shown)


christinebrass:
This looks good, and I can't think of any reason we wouldn't want the style loader stuff in the shared portion of the code. Thanks for this! :+1:

PR #62, opened 2016-02-11
Add a check for the extensions property

supergibbs:
Now the user will get a warning instead of a `property 'extensions' undefined` error


toddw:
:+1:

PR #63, opened 2016-02-11
Fix generator bug, create folders if missing

toddw:
I used `mkdirp` instead of `fs.mkdir` since it will ignore folders that already exist.


christinebrass:
:+1:

PR #64, opened 2016-02-11
Add --debug, -D mode for server side debugging

toddw:
Use case: `gluestick start -D` or `gluestick start --debug`

Client side debugging is much easier because you can use the browser's
developer tools. In a universal app, sometimes you will need to debug
code that will be rendered on the server. `node-inspector` is a great
tool for letting you debug server side code in Blink web developer
tools. It's a little slow to start up but it can really save you a lot
of time debugging server side code.

[Node Inspector](https://github.com/node-inspector/node-inspector)

screen shot 2016-02-11 at 3 02 57 pm

### DIFF
--- a/new/package.json
+++ b/new/package.json
@@ -33,6 +33,7 @@
     "babel-plugin-react-transform": "1.1.1",
     "eslint": "^1.10.3",
     "eslint-plugin-react": "^3.15.0",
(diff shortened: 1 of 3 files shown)


christinebrass:
:+1:

PR #65, opened 2016-02-12
Expose specified environment variables to client

toddw:
The application config file wasn't very useful because passing
environment variables into it made it so that it could only work on the
server side. If we wanted the environment variables to work on the
client side then webpack would need to know to expose them. **You
definitely don't want to just expose all environment variables or you
could leak things like api secrets**. It would also be annoying to have
to make an array of the variables that you wanted to expose.

To solve the problem, I decided to run static code analysis over the
config/application.js file to determine which environment variables the
developer was trying to use in the config/application.js file. I use
that to tell webpack which environment variables it needs to expose.

This was almost a complete solution until I realized that importing
config variables from this file in your application code would return
the entire object. So new projects will be generated with a slightly
modified application configuration file. I have also added an auto
upgrade script for backwards compatibility on all of the previously
generated applications.

### DIFF
--- a/src/update-config.js
+++ b/src/update-config.js
@@ -29,7 +29,7 @@ export default function updateConfig () {
       name: "confirm",
       message: `${chalk.red("The format of src/config/application.js is out of date. You should export the correct config object based on the environment.")}
 ${chalk.yellow("Example:")}\n${chalk.cyan(exampleContents)}
-Would you like to try to automaticaly update it?`
+Would you like to try to automatically update it?`
     };
     inquirer.prompt([question], function (answers) {
       if (!answers.confirm) return resolve();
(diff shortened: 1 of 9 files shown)


src/update-config.js
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    const exampleContents = fs.readFileSync(path.join(__dirname, "..", "new", "src", "config", "application.js"), "utf8");
+    const question = {
+      type: "confirm",
+      name: "confirm",
+      message: `${chalk.red("The format of src/config/application.js is out of date. You should export the correct config object based on the environment.")}
+${chalk.yellow("Example:")}\n${chalk.cyan(exampleContents)}
+Would you like to try to automaticaly update it?`
```
christinebrass: Small typo here

christinebrass:
Clever! :+1:

src/update-config.js
toddw: thanks for catching that! :smile_cat:

supergibbs:
Works on my machine :)

PR #66, opened 2016-02-12
CLI tests for the `generate` command: Part 1

christinebrass:
Adds a simple/small set of tests for the `gluestick generate` command. Another PR will be submitted for tests that validate that the appropriate files were generated for each type.


toddw:
:+1:

PR #68, opened 2016-02-13
CLI tests for the `generate` command: Part 2

christinebrass:
This PR adds tests that actually assert that files were generated/not generated under the right conditions.


toddw:
:+1:

PR #69, opened 2016-02-16
Clean up and standardize console.log calls

eitheror:
I tried to add a little bit of consistency to the various `console.log` calls in the app. To do so, I created a simple syntactic sugar wrapper around the `console.log(chalk.whatever())` calls. 
To use it, import the `lib/logger` and replace the console.log call with one of:
`logger.success` - green message
`logger.info` - yellow message
`logger.warn` - red message
`logger.error` - same as `warn` above, but prepends with the string "ERROR: ".

If you want to get fancy with your messaging, `logger` also exposes `_success`, `_info` and `_warn` which provide the same colorization without wrapping it in a `console.log` call. It also exposes a `_filename` method, which colorizes the text in cyan (and should be used when outputting file/path names) and a `_highlight` method, to be used when you want part of the string to be, well, highlighted.

example use:

```
import logger, {_highlight } from "./lib/logger";

logger.info(`This ${_highlight("really important")} message will be yellow`);
```


eitheror:
One note: pm2 appears to be stripping all colors from our console.log calls (regardless of the code in this pull request).

christinebrass:
With this change we may need to change spy code in tests that refer to `console.log` or `console.error`.

toddw:
Nice work! I was under the impression that the expected convention for variables names starting with an underscore meant they were to be considered private… Do you agree? If so, we shouldn't be exporting them or we should not use the underscored name. What do you think?

toddw:
> One note: pm2 appears to be stripping all colors from our console.log calls (regardless of the code in this pull request).

Do you get colors back if you remove `--raw` from this line:
https://github.com/TrueCar/gluestick/blob/develop/src/commands/start-server.js#L63

eitheror:
re method names: I share your concerns about using underscores, but I'm not sure I have a better solution. The main reasons for this change were:
-  Eliminate most of the direct calls to chalk, and apply a standard color scheme - this is handled by the underscore methods, e.g. `_success("text")` will always colorize "text" as green.
- Eliminate most of the `console.log` calls and make them more descriptive, which is handled by the non-underscore methods. i.e., when an operation is successful, we should use `logger.success("text")`. This will allow us to switch to a different, more robust logger if the need arises.

I toyed with a few other ways of exposing the closely correlated methods, such as using a unified method that takes an additional param (e.g. `.success("text", true)` to skip the `console.log` portion, or `.logInfo()` vs `.info()`, but they made the resulting code less readable in my opinion.

However, the more I think about it, the more I wonder if my initial approach of combining the decoration and the actual logging is faulty. Maybe they need to be two different modules, and something like this would make more sense?

```
//simplified colorScheme.js:

var chalk = require("chalk");

const info = chalk.yellow;
const highlight = chalk.bold;

module.exports = {
  info, highlight
};
```

```
//simplified logger.js
import colorScheme from "./colorScheme";

function info(text){
  console.log(colorScheme.info(text));
}

module.exports = {
  info
};
```

```
//simple consumer.js

import logger from "./lib/logger";
import { highlight } from "./lib/colorScheme";

logger.info(`This ${highlight("really important")} message will be yellow`);
```

Your thoughts?

eitheror:
As for pm2, removing `--raw` doesn't apply our colors, though it appears to add (and colorize) more data about the pm2 instance.

toddw:
@eitheror I really like your proposal for the alternative approach. It looks nice and it addresses my concern.

eitheror:
@toddw please take a look at the updated code.

toddw:
:+1: Please resolve the merge conflicts so I can merge

eitheror:
Done.

PR #70, opened 2016-02-19
Have webpack ignore server files

toddw:
We needed a way to have a require statement that only works on the
server side but we don't want webpack to try to bundle it.

Example:

```
if (typeof window === "undefined") {
  // webpack will ignore this line. The code in this conditional block
  // is not relevant to the browser side code anyways.
  var serverRequire = require("./require.server");
  serverRequire("./serverVersionOfFile");
}
else {
  require("./clientVersionOfFile");
}
```


christinebrass:
:+1:

PR #71, opened 2016-02-19
improve chunk names, improve proxy headers

toddw:
1. This gives us the ability to see something from the original names
   in code splits when the chunks are created.
2. `preserveHostHdr` forwards the original host through to the proxy in
   dev mode. This makes it easier to see where a request originated from
   when hitting the development server side rendering through the front-end
   proxy.


christinebrass:
:+1:

PR #74, opened 2016-02-25
Add flow header to generated files and update containers

toddw:
Added `/* @flow */` to generated files to better support flow.

While in there I noticed that containers/HomeApp.js and
containers/NoMatchApp.js no longer matched the output that `gluestick
generate container …` creates. So I updated those two files while I was
in there


christinebrass:
:+1:

PR #75, opened 2016-02-25
Add coverage reporting

christinebrass:
1) Get coverage working with the `test` command. For now, we are generating
html reports.

2) Extract karma config to separate file. I initially did this work as
part of creating a separate coverage command. I was going to revert it
but it seemed a little cleaner having the configuration separated from
the actual test command itself, so I left it as-is.

screen shot 2016-02-24 at 10 40 13 pm


toddw:
Looks great! Do we need to add anything to the .gitignore for the coverage reports?

christinebrass:
I thought we had to, but then I noticed that we already took care of that here: https://github.com/TrueCar/gluestick/blob/develop/new/_gitignore#L17-L18

christinebrass:
Also I forgot to mention that if we want to generate coverage for pre-existing gluestick apps, we'll need to run `gluestick start` first before `gluestick test` to get the updated node module install for isparta-loader. That's because I believe the autoupdate only currently runs for the start command. We might want to change that if it becomes a problem.

toddw:
:+1: I'm very very stoked about this! Thanks!

> if we want to generate coverage for pre-existing gluestick apps, we'll need to run gluestick start first before gluestick test

No problem at all!

PR #76, opened 2016-02-25
Updating packages and logic to handle react-router 2.0.0. Note: This …

mishkinf:
…change depends on corresponding gluestick-shared changes as well so each should be merged in lock-step. Please refer to the react router 2.0 upgrade guide (https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md).


toddw:
:+1:

PR #77, opened 2016-03-01
Upgrade to Babel 6

christinebrass:
**Please do not merge until https://github.com/TrueCar/gluestick-shared/pull/3 has been merged and versioned** since we'll need to update the version of gluestick-shared that his branch points to.

Summary of the work done:
- Pull in babel presets that are now required (stage-0, es2015, react)
- Use `babel-traverse` rather than `estraverse` since the latter didn't understand newer node types, even though we were using the latest version of it
- Fix imports where there are a mix of ES6 `export default` and CommonJS `require` imports by calling out the default using `.default`. See Kent Dodd's medium article below for more information. We can probably also change ES6 exports to CommonJS exports using `module.exports` if the use of `.default` everywhere looks ugly

---

**This is a breaking change**, in that already-existing gluestick apps will need to perform some manual updates to work well with this version of gluestick.

Update steps for already-existing apps:
1. Run `gluestick start` and answer 'Y' to update project dependencies. This will eventually start the gluestick server but it will be broken due to an oudated .babelrc file
2. ~~Update your project's .babelrc file, using this as a reference: https://github.com/TrueCar/gluestick/blob/develop/new/.babelrc~~ Answer 'Y' to update your .babelrc file automatically, or make sure to manually update to a compatible version.
3. Make sure containers are exported using the new format: https://github.com/TrueCar/gluestick/blob/develop/new/src/containers/HomeApp.js#L32-L35

---

What changed in Babel 6:
- https://medium.com/@malyw/how-to-update-babel-5-x-6-x-d828c230ec53#.q30v4mrpk
- http://jamesknelson.com/the-six-things-you-need-to-know-about-babel-6/
- https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0#.6twk0ifv2


toddw:
:+1:

PR #79, opened 2016-03-03
Feature/sela 227

JoeCortopassi:
added a version update based on the last version of Gluestick to modify the project. Version gets stored in the root `.gluestick` file, and gives an upgrade warning if the version is unknown or if your version of gluestick is different than the project's


package.json
```diff
@@ -73,11 +73,13 @@
     "react-transform-catch-errors": "1.0.0",
     "react-transform-hmr": "1.0.1",
     "redbox-react": "1.2.0",
+    "redux": "3.3.1",
     "rimraf": "2.5.1",
     "secure-handlebars": "1.2.0",
     "serialize-javascript": "1.1.2",
     "sha1": "1.1.1",
     "sinon": "1.17.2",
+    "temp": "0.8.3",
```
toddw: Does this qualify as a `devDependency`? is it only used for tests? if so we can move it to `devDependencies` and in the future if you want that to happen automatically `npm install _insert_name_ --save-dev`

src/cli.js
```diff
@@ -13,10 +13,7 @@ const generate = lazyMethodRequire("./commands/generate");
 const destroy = lazyMethodRequire("./commands/destroy");
 const dockerize = lazyMethodRequire("./commands/dockerize");
 
-const logger = require("./lib/logger");
```
toddw, line 16: Looks like your branch might be killing the work Or did to clean up our usage of `chalk` into a logging helper. Probably just a merge conflict but we should use the `logger` stuff Or added

src/cli.js
```diff
@@ -156,3 +169,38 @@ async function upgradeAndDockerize (name) {
   await autoUpgrade();
   dockerize(name);
 }
+
+function updateLastVersionUsed() {
+  const fileHeader = "DO NOT MODIFY";
```
toddw: Let's strip this header. The reason is, if anyone else wants to read the version from this file, which is why it is being created, they will have to strip it too. It won't be valid json otherwise.

src/cli.js
```diff
@@ -156,3 +169,38 @@ async function upgradeAndDockerize (name) {
...
+  fs.writeFileSync(gluestickDotFile, newContents);
+} 
+
+
+
+function shouldUpdateForProjectGluestickVersion(projectGluestickVersion) {
+  var mine = getVersion().split('.');
+  var project = projectGluestickVersion.split('.');
+
+  if (mine[0] !== project[0]
```
toddw: Is there a reason for splitting on the "." instead of just comparing `getVersion() === projectGluestickVersion` ?

src/commands/generate.js
```diff
@@ -1,9 +1,7 @@
 var fs = require("fs");
```
toddw, line 1: We want Or's changes back in this file

package.json
JoeCortopassi: Good catch. Not sure how that snuck in there :+1:

src/cli.js
JoeCortopassi: :+1:

src/cli.js
JoeCortopassi: didn't think of that :disappointed:. Good call :+1:

new/package.json
```diff
@@ -24,7 +24,7 @@
     "react-dom": "0.14.6",
     "react-redux": "4.0.6",
     "react-router": "2.0.0",
-    "redux": "3.0.5",
+    "redux": "3.3.1",
```
mishkinf: Is this just for fun or does it have to do with the changeset?

src/lib/updateVersion.js
```diff
@@ -0,0 +1,23 @@
...
+  const gluestickDotFile = path.join(process.cwd(), ".gluestick");
+  var fileContents = fs.readFileSync(gluestickDotFile, {encoding: "utf8"})
+  // We won't know how old this version is, it might have still have the 'DO NOT MODIFY' header
+  fileContents = fileContents.replace("DO NOT MODIFY", "");
+
+  // If the dot file only had the header (and possibly a new line character), then it definitely doesn't have the version check and is old
+  fileContents = (fileContents.length <= 1)? '{"version": "`unknown version`"}': fileContents;
+  var json = JSON.parse(JSON.stringify(fileContents));
+  if (getVersion() !== json.version) {
+    console.log(chalk.yellow("This project is configured to work with versions >= " + json.version + " Please upgrade your global `gluestick` module with `sudo npm install gluestick -g"));
```
eitheror: this should be using the logger instead of a direct call to console and chalk.

src/cli.js
```diff
@@ -58,29 +68,35 @@ commander
...
-  .action((options) => startAll(options.no_tests, options.debug));
+  .action((options) => startAll(options.no_tests, options.debug))
+  .action((options)=> updateLastVersionUsed());
 
 commander
   .command("build")
   .description("create production asset build")
-  .action(() => startClient(true));
+  .action(() => startClient(true))
+  .action((options)=> updateLastVersionUsed());
```
toddw: I believe eslint is going to complain that `options` is not used. Since we aren't using it we don't need to define the arg. This goes for all of the similar lines like this below

src/lib/updateVersion.js
```diff
@@ -0,0 +1,23 @@
...
+const path = require("path");
+const getVersion = require("./getVersion.js");
+const chalk = require("chalk");
+
+module.exports = function updateLastVersionUsed() {
+  // Check version in .gluestick file
+  const gluestickDotFile = path.join(process.cwd(), ".gluestick");
+  var fileContents = fs.readFileSync(gluestickDotFile, {encoding: "utf8"})
+  // We won't know how old this version is, it might have still have the 'DO NOT MODIFY' header
+  fileContents = fileContents.replace("DO NOT MODIFY", "");
```
toddw, line 11: I thought we decided not to do the whole `DO NOT MODIFY` line on the top. The reason is, it would be nice if scripts could just parse this file as JSON without having to strip the header. What do you think?

JoeCortopassi, line 11: This is just a safety check to strip it because we don't know how old the dot file is, and it might have one. We don't add it back in

src/cli.js
JoeCortopassi: :+1:

src/lib/updateVersion.js
JoeCortopassi: :+1:

src/lib/updateVersion.js
toddw, line 11: oh okay, cool got it :+1:

toddw:
:+1:

PR #85, opened 2016-03-08
update eslint-plugin-react version

eitheror:
eslint crashes with the old version of eslint-plugin-react when it tries to load the recommended settings from the eslintrc file. This fixes it.


toddw:
:+1:

PR #86, opened 2016-03-08
eval is faster than inline-source-map

toddw:


toddw:
Closing while we test this

PR #89, opened 2016-03-09
Fix css assets in dev and prod

toddw:
1. In development, we were only loading .css files but we want to
   support .scss, .sass as well… This adds support for that.
2. In production, we use an UglifyJS webpack plugin in our bundle. For
   some reason, the last published version of `gluestick-shared` had ES6+
   code in it which was causing the Uglify plugin to fail.
3. We were not reporting build errors so the failed build was failing
   silently.

### DIFF
--- a/src/commands/start-client.js
+++ b/src/commands/start-client.js
@@ -137,7 +137,15 @@ module.exports = function (buildOnly) {
   }
   else {
     logger.info("Bundling assets…");
-    compiler.run(() => {
+    compiler.run((error, stats) => {
+      const errors = stats.toJson().errors;
+      if (errors.length) {
+        errors.forEach((e) => {
+          logger.error(e);
+        });
+        return;
+      }
+
       logger.success(`Assets have been prepared for production.`);
       logger.success(`Assets can be served from the ${logsColorScheme.filename("/assets")} route but it is recommended that you serve the generated ${logsColorScheme.filename("build")} folder from a Content Delivery Network`);
     });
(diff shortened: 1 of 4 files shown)


src/commands/start-client.js
```diff
@@ -137,7 +137,15 @@ module.exports = function (buildOnly) {
   }
   else {
     logger.info("Bundling assets…");
-    compiler.run(() => {
+    compiler.run((error, stats) => {
```
christinebrass, line 140: Do we need to also handle fatal errors (`error` in this case)?

src/commands/start-client.js
```diff
@@ -137,7 +137,15 @@ module.exports = function (buildOnly) {
...
+      const errors = stats.toJson().errors;
+      if (errors.length) {
+        errors.forEach((e) => {
+          logger.error(e);
+        });
+        return;
+      }
+
```
christinebrass, line 148: I'm not sure under what conditions we'd get warnings, but maybe we'd want to logger.warn `stats.toJson().warnings` as well? Just thinking it could help us troubleshoot issues in the future, but if you think that's overkill I'm fine with what we have for now.

toddw, line 148: Let's come back to that in another ticket since this is breaking prod. I want to get it in quickly

src/commands/start-client.js
toddw, line 140: most likely yes, I just wanted to rush a fix in since production builds are currently broken

christinebrass:
Ok :+1:

PR #90, opened 2016-03-09
Fix bug in store

toddw:
We do not want/need the `.default` property on the reducers require
statement. Having it there breaks the store


christinebrass:
:+1:

PR #91, opened 2016-03-09
Better CLI Error Reporting

christinebrass:
This PR addresses the issue encountered when running gluestick commands outside of a gluestick-generated project. Some commands checked for the existence of a .gluestick file and reported an error, while others did not. This commonizes the functionality and ensures it is used where appropriate (excluding `gluestick new` of course).
- extract .gluestick file check into separate utility function
- create utility function for quitting program when not in a gluestick project
- use the utility function where appropriate
- add tests for newly used functions


toddw:
:+1:

PR #94, opened 2016-03-10
Clean up and improve generated tests

toddw:
We want our generated tests from `new` and `generate` to follow the same
pattern. They have gotten inconsistent.
We want to use enzyme in all of the generated tests.
We will now use `wrapper` instead of `renderedSubject`.
We are currently missing tests for HomeApp, NoMatchApp, and MasterLayout

### DIFF
--- a/test/commands/new.test.js
+++ b/test/commands/new.test.js
@@ -80,7 +80,7 @@ describe("cli: gluestick new", function () {
       // written for them. This will help us catch if we add a component or
       // container but we do not add a test.
       generatedFiles.forEach((file) => {
-        if (file.startsWith("src/components/") || file.startsWith("src/containers/")) {
+        if (/^src\/(components|containers)/.test(file)) {
           let testFilename = file.replace(/^src/, "test").replace(/\.js$/, ".test.js");
           expect(index[testFilename]).to.be.true;
         }
--- a/test/commands/new.test.js
+++ b/test/commands/new.test.js
@@ -62,4 +62,33 @@ describe("cli: gluestick new", function () {
     }
   });
 
+  it("should generate a test for all of the initial components and containers", () => {
+    const fakeNpm = sinon.stub(npmDependencies, "install");
+    try {
+      newApp("gs-new-test"); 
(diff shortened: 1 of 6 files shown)


test/commands/new.test.js
```diff
@@ -62,4 +62,33 @@ describe("cli: gluestick new", function () {
     }
   });
 
+  it("should generate a test for all of the initial components and containers", () => {
+    const fakeNpm = sinon.stub(npmDependencies, "install");
+    try {
+      newApp("gs-new-test"); 
+
+      // account for the fact that the gitignore file that gets renamed
```
christinebrass: ^Does this comment apply here?

test/commands/new.test.js
```diff
@@ -62,4 +62,33 @@ describe("cli: gluestick new", function () {
     }
   });
 
+  it("should generate a test for all of the initial components and containers", () => {
+    const fakeNpm = sinon.stub(npmDependencies, "install");
```
JoeCortopassi: why not have this and the restore in the `beforeEach` and `afterEach` you already have?

christinebrass:
Looks good aside from super minor comment :+1:

test/commands/new.test.js
```diff
@@ -62,4 +62,33 @@ describe("cli: gluestick new", function () {
...
+      const index = {};
+      generatedFiles.forEach((file) => {
+        index[file] = true;
+      });
+
+      // loop through and make sure components and containers all have tests
+      // written for them. This will help us catch if we add a component or
+      // container but we do not add a test.
+      generatedFiles.forEach((file) => {
+        if (file.startsWith("src/components/") || file.startsWith("src/containers/")) {
```
toddw: This would work better as something like `/^src\/(components|containers)/` updating…

test/commands/new.test.js
toddw: nope, copy/paste error… :+1: fixing

JoeCortopassi:
:+1:

PR #96, opened 2016-03-11
Cover reports for Gluestick

christinebrass:
This PR adds cover reports for the Gluestick tool itself. We are using babel-istanbul here to get reports generated in their original formats. 

Changes:
- Remove the need to require babel-polyfill in individual tests, and move it to the require hook on the test command itself
- Add `test-coverage` command to generate reports
- Move version update test to reflect actual src structure (cleanup)


toddw:
:+1:

PR #97, opened 2016-03-14
Lint, cleanup and add some tests

christinebrass:
This PR consists of some edits made while browsing through the code. I noticed coverage was still low for the destroy command, so added more tests. It can still use more which we will add as we go.
- Lint and cleanup code
- Add a couple more destroy command tests


JoeCortopassi:
:+1:

PR #98, opened 2016-03-14
Better app coverage reports

christinebrass:
Using babel-istanbul reconciles babel output against sourcemaps, providing apps with better coverage reports.

We use this for Gluesick itself, now let's use it for Gluestick-generated apps.


toddw:
:+1:

PR #100, opened 2016-03-15
Sorting packages alphabetically to resolve issue with merge conflicts

mishkinf:


src/auto-upgrade/updatePackage.js
```diff
@@ -12,6 +12,23 @@ const PROJECT_PACKAGE_LOCATION = path.join(process.cwd(), "package.json");
...
+/**
+ * This function is used to put the dependencies in alphabetical order.
+ * It takes a hash and returns the same hash ordered by its keys.
+ *
+ * @param {Object} Unordered hash which one would like to be sorted by keys
+ *
+ * @return {Object}
+ */
+function orderHashByKeys (unordered) {
+  const ordered = {};
```
toddw: Unfortunately this will not work… The spec for objects in Ecmascript state:

> ### 4.3.3 Object
> 
> An object is a member of the type Object. It is an **unordered collection of properties** each of which contains a primitive value, object, or function. A function stored in a property of an object is called a method.

[Old ES3 Spec where javascript objects are defined](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%203rd%20edition,%20December%201999.pdf)

In my experience, Chrome actually does jumble the order. I'm sure it's for optimization reasons.

That being said, we need to use something else. ES2015 gives us the Map datatype. This is what you want to use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

From there though, you will need to see what happens when you JSON.stringify a map, I'm actually not sure. It might just work, but it might require some more work on your part.

toddw: So you can't stringify Map objects… I looked into what `npm` uses and they use an module called `sorted-object` https://www.npmjs.com/package/sorted-object

It works really well.

Try this sample code in the following url:
https://tonicdev.com/npm/sorted-object

```
var sO = require("sorted-object");
var x = {
  a: "Todd",
  z: "Zeke",
  b: "Ambre"
};
var y = sO(x);
JSON.stringify(y);
```

toddw: https://tonicdev.com/56e847786b502a11001a3162/56e847fe91193c1100ca8589

src/auto-upgrade/updatePackage.js
```diff
@@ -167,9 +184,12 @@ function performModulesUpdate (mismatchedModules, done) {
   let module;
   for (let moduleName in mismatchedModules) {
     module = mismatchedModules[moduleName];
-    projectPackageData[module.type][moduleName] = module.required
+    projectPackageData[module.type][moduleName] = module.required;
```
toddw, line 171: :+1:

toddw:
Can we write a test for this?

mishkinf:
We can write tests for this but it's more complicated than necessary to do so. Requires a fair amount of mocking and deep understanding of the code which makes it less than ideal and highly likely to be a little frustrating if we need to make changes. I actually have code that tests some of this file. I want to discuss ways to restructure the code to make it more testable though. Seems more ideal than me forcing the issue with annoying to write / modify tests. Let me get some of these thoughts more formalized.

toddw:
:+1:

PR #103, opened 2016-03-17
Fix version update

christinebrass:
This PR addresses the minor issue where a warning is displayed to the user when his gluestick version is more recent than the project version. That's a normal case, so we don't need to display the warning. 

I also took the opportunity to lint the files and refactor for easier testing. Specifically, we are now passing the gluestick version to the update function so we can test behaviors for different versions.


toddw:
:+1:

PR #105, opened 2016-03-18
Improve font support

christinebrass:
This PR addresses the issue of not being able to properly load font-awesome assets, as some font urls have an additional query parameter with version information appended to them. Example:

```
url(../fonts/myfont.woff?v=1.1.1). 
```
- Upgrade webpack-isomorphic-tools to latest version to get regular expression support
- Add regular expression to Gluestick's configuration file support fonts with versions
- Add the same for images, as I was seeing a lingering issue with svg's that have versions on them. **Side note**: I'm unsure if this should _only_ apply to svg files or if this should apply to all image types. I made it support all for now but we can change this if that doesn't seem right.
- Lint

To test this, I created a sample repo and made sure it loaded assets with `gluestick start`: https://github.com/cybrass/test-font-awesome-gs

### DIFF
--- a/src/lib/webpack-isomorphic-tools-configuration.js
+++ b/src/lib/webpack-isomorphic-tools-configuration.js
@@ -36,7 +36,8 @@ let userExtensions = [];
 module.exports = {
   assets: {
     images: {
-      extensions: ["png", "jpg", "gif", "ico", "svg"]
+      extensions: ["png", "jpg", "gif", "ico", "svg"],
+      regular_expression: /\.(png|jpg|gif|ico|svg)(\?v=\d+\.\d+\.\d+)?$/
     },
     styles: {
       extensions: ["css", "scss", "sass"],
--- a/src/lib/webpack-isomorphic-tools-configuration.js
+++ b/src/lib/webpack-isomorphic-tools-configuration.js
@@ -1,9 +1,9 @@
 import logger from "./logger";
 import { highlight } from "./logsColorScheme";
 
-var path = require("path");
-var process = require("process");
-var WebpackIsomorphicToolsPlugin = require("webpack-isomorphic-tools/plugin");
+const path = require("path");
+const process = require("process");
+const WebpackIsomorphicToolsPlugin = require("webpack-isomorphic-tools/plugin");
 
 // We use `get-webpack-additions` to get the data back because it will first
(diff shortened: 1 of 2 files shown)


christinebrass:
Here's the svg issue seen (causing me to add the regex to images as well):

```
ERROR in ./assets/fonts/fontawesome-webfont.svg?v=4.5.0
Module parse failed: /opt/truecar/src/fatest-gs/assets/fonts/fontawesome-webfont.svg?v=4.5.0 Line 1: Unexpected token <
You may need an appropriate loader to handle this file type.
| <?xml version="1.0" standalone="no"?>
| <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" >
| <svg xmlns="http://www.w3.org/2000/svg">
 @ ./~/css-loader!./~/sass-loader!./assets/css/font-awesome.min.css 6:690-741
```

src/lib/webpack-isomorphic-tools-configuration.js
```diff
@@ -36,13 +36,14 @@ let userExtensions = [];
 module.exports = {
   assets: {
     images: {
-      extensions: ["png", "jpg", "gif", "ico", "svg"]
+      extensions: ["png", "jpg", "gif", "ico", "svg"],
+      regular_expression: /\.(png|jpg|gif|ico|svg)(\?v=\d+\.\d+\.\d+)?$/
```
toddw, line 40: Do you think this regex is better than `/\.(png|jpg|gif|ico|svg)(\?v=\d+\.\d+\.\d+)(\?.*)?$/`? Yours will work for font awesome but what if another library has something else after the question mark? Maybe mine is too broad, what do you think?

src/lib/webpack-isomorphic-tools-configuration.js
```diff
@@ -50,6 +51,7 @@ module.exports = {
     },
     fonts: {
       extensions: ["woff", "woff2", "ttf", "eot"],
+      regular_expression: /\.(woff|woff2|ttf|eot)(\?v=\d+\.\d+\.\d+)?$/,
```
toddw, line 54: same comment as above

toddw:
:+1:

PR #111, opened 2016-03-22
Add option to run test suite only once

christinebrass:
By default we run app tests in continuous mode; add an option to exit after running the test suite once.


JoeCortopassi:
:+1:

PR #112, opened 2016-03-22
change devtool from `eval` to `cheap-module-eval-source-map`

toddw:
`eval` doesn't give us the original code `cheap-module-eval-source-map`
does


JoeCortopassi:
:+1:

PR #115, opened 2016-03-25
update radium version

eitheror:
Version 0.17.0 allows us to set radium into test mode, which prevents the annoying "addCSS" errors on tests when using breakpoints in your component.


toddw:
:+1:

PR #118, opened 2016-03-28
Make sure to perform a numeric comparison

christinebrass:
Before this update, we were performing string comparison of version numbers. This means that a gluestick version of "0.0.10" will compute to be older than a version of "0.0.8" which is incorrect.


toddw:
:+1:

PR #120, opened 2016-03-31
Generate cobertura reports for CI

christinebrass:
Eventually we want to parameterize these, so users can generate whatever flavors of reports works for them. For now, I'd like to create cobertura reports for integration with our build servers.

Coverage directory after running `gluestick test`:

```
coverage
├── html
│   └── index.html
│   └── ...
└── xml
    └── cobertura-coverage.xml
```


JoeCortopassi:
:+1:

PR #122, opened 2016-04-05
Make it so routes returns function instead of object

toddw:
React-router allows us to do some pretty cool async stuff with route
properties like `onEnter`, `getComponent`, and `getComponents`. When we
were simply returning an object, we didn't have access to the redux
store from so it was limiting what we could accomplish in those async
methods.

By converting `routes` to a method that gets the redux store as the
first argument, it lets us do things like wait for actions to dispatch
before rendering a route. This makes things like fetching initial data
or even authentication much easier to work with.

Converting the routes object to a method was not going to be easy since
the amount of code in that object could potentially be very complex and
it isn't safe to assume we can just replace the export method and the
last closing part of that tag with a function, then add a return
statement and tab everything over. Instead, a deprecation notice was
added so that developers can handle this on their own.

Lastly, the `match` call we were performing in `.entry` was incorrect,
it was not actually matching on the routes it should but this PR fixes
that.


christinebrass:
👍

PR #123, opened 2016-04-07
update *all* of the packages

toddw:
We need to stay up on this, since we haven't been running `npm outdated`
and keeping up to date this is a much riskier update.

screen shot 2016-04-07 at 3 56 44 pm
- [x]  All unit tests pass
- [x] Manually tested apps generated with GlueStick after update


christinebrass:
👍 🍊

PR #125, opened 2016-04-19
Integrate react-helmet for document head changes

christinebrass:
This PR integrates react-helmet into new Gluestick projects and allows a base set of document head tags to be configured via the application's config file:
https://github.com/cybrass/helmet-gluestick/blob/master/src/config/application.js#L4-L10 

I also added react-helmet to generated containers by default so users of Gluestick could immediately take advantage of it: 
https://github.com/cybrass/helmet-gluestick/blob/master/src/containers/Mycontainer.js#L27

Full working example here: https://github.com/cybrass/helmet-gluestick

### DIFF
--- a/generate/container.js
+++ b/generate/container.js
@@ -2,6 +2,7 @@
 import React, { Component } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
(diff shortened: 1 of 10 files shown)


toddw:
:+1:

PR #127, opened 2016-05-02
Linting and minor refactoring

christinebrass:
This PR consists of some re-organization and cleanup I did while going through the modules. 
- Use more consistent module naming conventions
- Lint src, test and template directories and add a script for it (`npm run lint`)
- Consolidate configuration into one directory so we know what is configuration vs. implementation code
- Enhance the logging functionality by allowing multiple args to be used with `console.log` statements and refactor any `console.log` statements to using the logger


src/commands/dockerize.js
```diff
@@ -1,10 +1,9 @@
 import { spawn } from "cross-spawn";
 import path from "path";
 import process from "process";
-import fs from "fs";
 
 module.exports = function (name) {
-  const { version: gluestickVersion } = JSON.parse(fs.readFileSync(path.join(__dirname, "../../package.json"), "utf-8"));
+  // @TODO: check if docker is installed first
```
toddw: Can we get a reference to a ticket for this?

src/lib/detectEnvironmentVariables.js
```diff
@@ -1,26 +1,30 @@
...
-var traverse = require("babel-traverse").default;
+const babel = require("babel-core");
+const traverse = require("babel-traverse").default;
+const logger = require("./logger");
 
 /**
  * Read the contents of a file, convert it to an abstract syntax tree and
  * figure out how many process.env.X properties are accessed. Return an array
- * of environment variables.
+ * of environment constiables.
```
toddw: constants? 😸

src/lib/server/index.js
```diff
@@ -0,0 +1,31 @@
+import express from "express";
+import logger from "../logger";
+import { filename } from "../logsColorScheme";
+import requestHandler from "./requestHandler";
+
+const isProduction = process.env.NODE_ENV === "production";
+//@TODO: read host from config file
```
toddw: ticket please :smile:

src/lib/server/index.js
```diff
@@ -0,0 +1,31 @@
...
+const host = "localhost";
+//@TODO: understand why different port numbers between prod and dev
```
toddw: In development mode, we have webpack-dev-tools serving assets on one port and express serving the server side rendering on another port. In development mode, we proxy requests to the webpack dev tools server so that you don't have to point your code to two different ports. In production we are not proxying to another server, but instead serving the static assets ourselves.

https://github.com/TrueCar/gluestick/blob/develop/src/commands/start-client.js#L114

If I remember correctly, a request comes into the express app using the webpack-dev-server. If it finds the asset, then it returns the asset, if not then if tries to pass it along to the server side rendering port. So in development mode, you access the app on the webpack dev server route.

In production, we actually mount /assets to point to the pre-bundled assets.

I'm fairly certain this can be done better.

src/lib/server/index.js
```diff
@@ -0,0 +1,31 @@
...
+//@TODO: read port number from config file
```
toddw: ticket please 😸

src/lib/detectEnvironmentVariables.js
christinebrass: Lol, global replace error. I'll fix that.

src/commands/dockerize.js
christinebrass: https://github.com/TrueCar/gluestick/issues/128

src/lib/server/index.js
christinebrass: https://github.com/TrueCar/gluestick/issues/129

src/lib/server/index.js
christinebrass: Awesome, thanks for the explanation @toddw! I'll remove my todo comment for this now. This was actually an internal comment to myself that I accidentally left, but glad I got something out of it 😄

src/lib/server/index.js
christinebrass: https://github.com/TrueCar/gluestick/issues/129

toddw:
:+1:

PR #131, opened 2016-05-03
SELA-301 Support Mocha without Karma as a test runner.

threehams:
Set up Mocha to run tests without needing to spin up a browser or bundle. This is very close to a drop-in replacement right now.
### Changes
- Assets (scss, svg, png) need to change to use relative paths. `require.extensions` only works for file imports, not for modules, and anything without a relative path is assumed to be a module. This means that anything using a path such as `import ribbonUrl from "assets/img/icons/ribbon.svg"` will pass in the Karma tests, and work in the app, but fail tests using Mocha.

I'm trying to find a way to get this working, but the directory aliasing is done through Webpack, and is not based on any standard.
#### Mocha + Karma + Webpack

Startup time: 43 seconds
Time from file save to results: 10-11 seconds
#### Mocha + Node

Startup time: 4 seconds
Time from file save to results: 400-800 ms

### DIFF
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -3,7 +3,6 @@ const Server = karma.Server;
 const runner = karma.runner;
 const karmaConfig = require("../config/karma-config").default;
 const spawn = require("cross-spawn").spawn;
-const mocha = require("mocha");
 const path = require("path");
 const CWD = process.cwd();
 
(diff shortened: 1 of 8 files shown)


src/commands/test.js
```diff
@@ -2,23 +2,38 @@ const karma = require("karma");
...
+const spawn = require("cross-spawn").spawn;
+const mocha = require("mocha");
+const path = require("path");
+const CWD = process.cwd();
 
 module.exports = function (options) {
+  // override to run tests in Node without Karma/Webpack
+  if (options.mochaOnly) {
+    const helperPath = path.join(__dirname, "..", "lib", "testHelperMocha.js");
+    const testPath = `${CWD}/test/**/*.js`;
```
christinebrass: If we want to be consistent with our karma test helper, we'd use `*.test.js` as opposed to `*.js` at the end here: https://github.com/TrueCar/gluestick/blob/develop/src/lib/testHelper.js#L11

christinebrass:
👍

PR #139, opened 2016-05-05
All new projects should use v0.5.0 of the image

christinebrass:
Using GitHub's Release page doesn't ensure that this gets updated and committed. We will need to either remember to run `npm run prepublish` and commit the results before every release or write a different script for updating our npm version, which will ensure this step also happens.


christinebrass:
👍

PR #148, opened 2016-05-10
add support for isomorphic http client that forwards cookies

toddw:
# Depends on https://github.com/TrueCar/gluestick-shared/pull/9 being merged and new version of Gluestick-Shared being released first

> When doing HTTP requests from the browser, the browser's cookies are automatically included in the request. When doing HTTP requests from node, the browsers cookies will not be automatically included. To resolve this issue, we added proxy support to GlueStick, by putting API urls behind a proxy the host for the API and the client can be shared and thus share the same cookies. When we are going to be making an HTTP request from the node app, we simply forward all of the headers from the original browser request. This means we have a split where code run in the browser should use a plain `axios` client and the requests from the server need a custom instance of `axios` where the headers from the initial request are injected automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "author": "Todd Williams <toddsurfs@icloud.com>",
   "dependencies": {
     "app-module-path": "1.0.6",
+    "axios": "0.11.0",
     "babel-cli": "6.8.0",
     "babel-core": "6.8.0",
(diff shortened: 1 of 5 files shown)


toddw:
retest this please

christinebrass:
👍

PR #150, opened 2016-05-10
Add coverage reports, both in watch and single modes.

threehams:
Do lots of cleanup on test.js structure.


src/commands/test.js
```diff
@@ -6,60 +6,142 @@ const spawn = require("cross-spawn").spawn;
...
+    spawn(...spawnOpts.coverage);
+  }
+
+  function watch() {
+    // If not debugging, start up mocha in watch mode.
+    spawn(...spawnOpts.mocha);
+
+    // Set up a watcher to create coverage reports in the background, only messaging on success/failure.
+    // To ensure only the latest version is available, kill/restart the process on each change.
+    let mocha;
```
threehams: I don't like this variable name, but "process" is already taken. Suggestions?

src/commands/test.js
```diff
@@ -6,60 +6,142 @@ const spawn = require("cross-spawn").spawn;
...
-  const testPath = `${CWD}/test/**/*.test.js`;
-  const helperPath = path.join(__dirname, "..", "lib", "testHelperMocha.js");
+const MOCHA_PATH = path.join(__dirname, "..", "..", "node_modules", ".bin", "mocha");
+const MOCHA_PATH_COVERAGE = path.join(__dirname, "..", "..", "node_modules", ".bin", "_mocha");
+const MOCHA_ENV = { ...process.env, ...{ NODE_PATH: CWD, NODE_ENV: "test" }};
+const MOCHA_TEST_PATH = `${CWD}/test/**/*.test.js`;
+const MOCHA_HELPER_PATH = path.join(__dirname, "..", "lib", "testHelperMocha.js");
+const BABEL_NODE_PATH = path.join(__dirname, "..", "..", "node_modules", ".bin", "babel-node");
+const BABEL_ISTANBUL_PATH = path.join(__dirname, "..", "..", "node_modules", "babel-istanbul", "lib", "cli.js");
+const INSPECTOR_PATH = path.join(__dirname, "..", "..", "node_modules", ".bin", "node-inspector");
```
christinebrass: Since the node_modules path is used a lot, you can even separate that out if you think it will be cleaner:

```
const NODE_MODULES = path.join(__dirname, "..", "..", "node_modules");
const MOCHA_PATH = path.join(NODE_MODULES, ".bin", "mocha")
const MOCHA_PATH_COVERAGE = path.join(NODE_MODULES, ".bin", "_mocha");
const BABEL_NODE_PATH = path.join(NODE_MODULES, ".bin", "babel-node");
...
```

threehams: 👍

toddw:
:+1:

PR #151, opened 2016-05-11
Update enzyme dependency for new projects

christinebrass:
https://github.com/airbnb/enzyme/blob/master/CHANGELOG.md


toddw:
:+1:

PR #153, opened 2016-05-12
Swap out express-http-proxy for http-proxy-middleware

toddw:
I was running into some strange issues with express-http-proxy where it
was just throwing Socket Hung Up… After spinning my wheels for a while I
cam across `node-http-proxy` which has over 5k stars on GitHub and has
had a ton of community support. Finally, `http-proxy-middleware` is express
middleware that implements `node-http-proxy`.

After switching to the new module, things work great and I was able to
clean up some code we had before as well.


christinebrass:
👍

PR #154, opened 2016-05-12
Remove babel config check from auto-update utility

### LINKED ISSUE #146: Outdated .babelrc message
If a Gluestick app has updated its .babelrc file to include a new plugin (for example), a message will appear during `gluestick start` that their .babelrc file is outdated:

```
› gluestick start
...
? The format of .babelrc is out of date. Would you like to automatically update it? (Y/n)
```

Sometimes an app will want to use additional configuration options or plugins not specified here: https://github.com/TrueCar/gluestick/blob/develop/templates/new/.babelrc. We need to prevent the message from appearing in that case.

christinebrass:
Resolves #146

Users of Gluestick will likely want to add their own set of plugins and
presets into their applications. We shouldn't warn them if their babelrc
file differs from the base file produced when they first created the
project.


toddw:
:+1:

PR #157, opened 2016-05-12
Add webpack aliases for component and container folders

### LINKED ISSUE #155: Add webpack alias for components and containers
Just as we alias assets, it would be nice to also alias the components and containers folders.

christinebrass:
Also update template files so newly-generated Gluestick apps use the simplified path in imports.

Resolves #155


toddw:
:+1:

PR #158, opened 2016-05-12
Set Mocha/Node to be the default test runner.

threehams:
Override with --karma option.

Setting now that mocha/node is at feature parity with the existing Karma setup (watch, coverage, debugging).


toddw:
:+1:

PR #159, opened 2016-05-12
Constrain server watch to related files and folders.

threehams:
Prevent server from restarting on unrelated file changes, mostly 'coverage' and 'test'.


mishkinf:
👍

PR #161, opened 2016-05-13
update to use `getHttpClient` and pass config/httpConfig to it

toddw:
# Relies on https://github.com/TrueCar/gluestick-shared/pull/10 being merged and published first

We wanted to add support for customizing the httpClient with options in the application/config file. This update along with the updates to `gluestick-shared` make that possible. https://github.com/TrueCar/gluestick-shared/pull/10


christinebrass:
👍

PR #163, opened 2016-05-13
Update gluestick-shared dependency version

christinebrass:
Without this patch, Gluestick will re-prompt to update dependencies for each `gluestick start` command.


toddw:
:+1:

PR #166, opened 2016-05-16
Support directories in `gluestick destroy` arguments

### LINKED ISSUE #165: Allow removal of directories with the `gluestick destroy` command
We added the ability to create directories with `gluestick generate` in #156. Support the removal of those directories via the `gluestick destroy` command:

```
› gluestick destroy component common/Foobar
Removed file: /private/tmp/myproject/src/components/common/Foobar.js
Removed file:/private/tmp/myproject/test/components/common/Foobar.test.js
```

christinebrass:
Resolves #165 
- Parse the dirname and basename from the provided path and use that during file generation
- Add tests for the new functionality
- Organize all tests by grouping them by context

### DIFF
--- a/src/commands/destroy.js
+++ b/src/commands/destroy.js
@@ -20,28 +20,35 @@ module.exports = async function (command, name) {
     return;
   }
 
+  const dirname = path.dirname(name);
+  const basename = path.basename(name);
+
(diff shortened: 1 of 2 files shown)


toddw:
:+1:

Issue #168, opened 2016-05-17
Background coverage report generation is noisy.


threehams:
The background process currently inherits stderr, which for Mocha tests adds a lot of noise when running in watch mode. We could safely suppress this output.

christinebrass:
Resolved by #168

PR #169, opened 2016-05-17
#168 Suppress stderr on coverage report generation.

threehams:


christinebrass:
So we'll still see stderr during the actual test run, we just won't be redundant by showing it again during coverage report generation. 👍

PR #170, opened 2016-05-17
bump gluestick-shared version (resolves dependency issues)

toddw:
integrate changes from https://github.com/TrueCar/gluestick-shared/pull/12


JoeCortopassi:
👍

PR #171, opened 2016-05-18
Fix import path of generated tests

christinebrass:
Generated tests for containers and components were being generated with
the wrong import path to the container or component. Make sure the
correct relative path is being included in tests.


toddw:
:+1:

Issue #172, opened 2016-05-18
Optimize docker images


natefox:
After chatting with @joshuago we've come up with two optimizations and ideas for a third.  

The first is simply to merge the steps that `node:6` takes as well as the existing commands into one layer and then remove the installed files. This creates a layer that is quite a bit smaller than the multiple layers that are created when the commands are ran separately. This PR should be 100% compatible with the existing Dockerfile (unless the dev libraries are required for operation).
Original Dockerfile ~ 900Mb
Small Optimization Dockerfile ~ 710Mb

The second is to do the same as the first, but also runs an `apt-get autoremove` which removes any orphaned packages. This _may_ break things as it may remove required libraries. I havent ran all of the tests in the resulting docker container.
Original Dockerfile ~ 900Mb
Large Optimization Dockerfile ~ 465Mb

A third idea is to rebuild based on Alpine Linux. This would require some digging through the libraries from `node:6` to determine what is actually needed. At a cursory glance, it appears that some of the Debian libraries are not available in Alpine (or at least renamed).

Lastly, in both submitted Dockerfiles, I've moved the ENV lines to the top of the file. This has resulted in an additional ~50Mb savings. I believe the root cause is that `npm install` wasnt seeing these environment variables when they were being ran and therefore installing the devDependencies along with the production dependencies.

toddw:
Thanks @natefox this is great work!

cybrass:
This also looks great! I had to had the minor workaround I posted in #173 and once I did, things ran better. 👍

christinebrass:
Resolved by #174

PR #173, opened 2016-05-18
minor dockerfile space optimization

natefox:
See Issue #172 for more detail - this is the first/smaller of the two optimizations

PR #174, opened 2016-05-18
docker large optimization

natefox:
See Issue #172 for more detail - this is the second/larger of the two optimizations


christinebrass:
This also looks great! I had to had the minor workaround I posted in #173 and once I did, things ran better. 👍

cybrass:
Good to merge once develop is merged in.

PR #179, opened 2016-05-19
Add check for docker before dockerizing

cjeich:
Hey guys, awesome project.

Here's a patch to update check if docker exists before running the command.


toddw:
Thanks @cjeich! 😄

toddw:
Would you mind making a minor update to this? We added colors to our log messages and we try to use https://github.com/TrueCar/gluestick/blob/develop/src/lib/logger.js for logging messages.

toddw:
example in the wild: https://github.com/TrueCar/gluestick/blob/develop/src/lib/utils.js

christinebrass:
Yes, thanks for your contribution @cjeich! I was thinking it might be nice to use something like shelljs for this, which is a dependency we could use within other areas of Gluestick as well: https://github.com/shelljs/shelljs

For this particular case, we can simply use the `which` command: https://github.com/shelljs/shelljs#whichcommand

What do you guys think?

cjeich:
Easy enough to change. Here's the update.
Shelljs definitely makes sense if we need a unixy way to manipulate files in a cross platform way.

christinebrass:
Awesome, @cjeich, this looks great! We will be able to merge once the CLA we sent you is signed. Thanks again 😄

cjeich:
Hey, @cybrass I didn't get the CLA can you sent it again.
my email is my github username [at] gmail.com

christinebrass:
👍

PR #181, opened 2016-05-20
Prevent colors from rendering on server errors

christinebrass:
For our users who prefer lighter background schemes.


toddw:
:+1:

PR #182, opened 2016-05-20
add support for vendor bundle

toddw:
This allows us to specify which node_modules we want to be included in a vendor bundle by adding an array of module names in the webpack-additions file. `vendor: ["react", "react-router"]`

http://webpack.github.io/docs/code-splitting.html#split-app-and-vendor-code


christinebrass:
Looks good 👍

PR #184, opened 2016-05-23
Bundle Size Optimization Updates

toddw:
1. Add support for webpack plugins. This allows us to use things like webpack's `IgnorePlugin` plugin and/or the `ContextReplacementPlugin`. (http://stackoverflow.com/questions/25384360/how-to-prevent-moment-js-from-loading-locales-with-webpack)
2. To better understand the bundle size, the webpack stats are now output to webpack-bundle-stats.json This file can be piped to `webpack-bundle-size-analyzer` like `cat webpack-bundle-stats.json | webpack-bundle-size-analyzer` for more detailed results.
3. Made `react-router` imports more specific as to not include parts of the library we are not using.

### DIFF
--- a/src/commands/start-client.js
+++ b/src/commands/start-client.js
@@ -1,3 +1,4 @@
+const fs = require("fs");
 const path = require("path");
 const webpack = require("webpack");
 const process = require("process");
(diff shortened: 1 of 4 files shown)


christinebrass:
👍

PR #186, opened 2016-05-25
Improve production mode

toddw:
1. Perform a clean up before pm2 starts that will make sure there were
   no phantom processes running before kicking off new phones. (This one
   helps both production and development)
2. Run the gluestick build as part of creating a docker image. This way the
   server will be ready as soon as it starts. Added "skip build" option to
   arguments so that the docker image will stop kicking off a build when it
   starts. Kicking off a build is still useful when running locally and you
   haven't run `gluestick build` before running with NODE_ENV as
   production.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -66,5 +66,4 @@ RUN apt-get update && \
 
 EXPOSE 8888
 
-RUN gluestick build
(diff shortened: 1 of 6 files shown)


mishkinf:
👍

PR #187, opened 2016-05-25
Resolve Dockerfile issue and pm2 error

toddw:
`RUN gluestick build` was added to the wrong file
`templates/new/src/config/.Dockerfile` which is actually overwritten at
runtime by `docker/generate-dockerfile.js`. This commit will resolve
that.

There were instances where pm2 would throw an error if it tried to stop
an instance that it knew nothing about. For some reason this error
condition wasn't always repeatable. To resolve the issue I added a function that will return whether or not an instance for a given name exists and then use that to determine if we should attempt to stop or delete.


christinebrass:
👍

PR #196, opened 2016-06-04
Add support for multiple entrypoints

toddw:


src/lib/isChildPath.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  for (let i = 0; i < parentSplitLength; i++) {
+    if (parentSplit[i] !== childSplit[i]) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
christinebrass, line 9: Another thing we can do here is use `path.relative(parent, child)` and if the path starts with `.` then we know it isn't a parent.

toddw, line 9: nice! I like that, I'll do that

christinebrass:
👍

PR #199, opened 2016-06-11
Add Gluestick prefix to logging statements

christinebrass:
So that users know where the logging statements are originating from:

logger


webmanarmy:
This is great! Can we change it to `[GlueStick]`?

toddw:
This is great! Can we change it to [GlueStick]?

christinebrass:
Done! 😄

toddw:
:+1: Will merge as soon as build passes

PR #206, opened 2016-06-28
Improved server logging

christinebrass:
This PR adds [pino](https://github.com/mcollina/pino) for logging. This will give us JSON-formatted logs in production and pretty logs in development mode.

We want to keep the CLI logging format separate from the server logging, and the cliLogger now lives in a module of that name.


toddw:
:+1:

PR #209, opened 2016-06-29
Proxy error logging changes

christinebrass:
- Move request logging middleware up middleware chain so that proxy errors can be caught and logged
- Remove pretty-error now that we are using pino for logging


toddw:
:+1:

PR #211, opened 2016-07-07
Add support for `require.context`

toddw:
Webpack isomorphic tools has enabled the ability to use `require.context` but you have to flip on the `patch_require` flag in configuration to enable it.

https://github.com/halt-hammerzeit/webpack-isomorphic-tools#configuration

somewhat related discussion about require.context and webpack-isomorphic-tools: https://github.com/halt-hammerzeit/webpack-isomorphic-tools/issues/48


christinebrass:
👍

PR #212, opened 2016-07-07
Stop loading every single css file into the DOM

toddw:
We previously loaded every single css file into the DOM in development
mode to resolve the flash of unstyled content (FOUC). This turned out to
be a mistake because when you are dealing with a large app that has
multiple entry points or code splitting every single css file is
included. In one example this resulted in over 400k lines of CSS to be
loaded in the DOM.

We will create another ticket to tackle the FOUC in a better way.


christinebrass:
👍

PR #215, opened 2016-07-08
Use state to set response status codes

christinebrass:
Depends on merge of https://github.com/TrueCar/gluestick-shared/pull/22

If a GlueStick error handling action was dispatched, an error object will be available on the state that can be used to appropriately set the response status code.


toddw:
:+1:

PR #216, opened 2016-07-11
fix missing stack trace for some 500 errors

saltycrane:
In v0.8.6, for some server-rendered 500 errors, I don't see a stack trace when I expect to see one. For example, this code in Home.js from a newly created gluestick project causes a 500 error:

```
/* @flow */
import React, { Component } from "react";

export default class Home extends Component {
  render () {

    asdf;

    return (
      <div>Home</div>
    );
  }
}
```

Before this fix, the browser shows an empty box when hitting the page directly or reloading:

screen shot 2016-07-11 at 12 16 49 pm

After the fix, it shows the stack trace:

screen shot 2016-07-11 at 12 17 06 pm

### DIFF
--- a/src/lib/server/requestHandler.js
+++ b/src/lib/server/requestHandler.js
@@ -118,9 +118,9 @@ module.exports = async function (req, res) {
       }
       catch (error) {
         // Render any error that was set on GlueStick's internal state object if one exists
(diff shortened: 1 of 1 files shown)


christinebrass:
Thanks! 👍

PR #219, opened 2016-07-19
Update error status code retrieval

christinebrass:
Second part of cleanup done here: https://github.com/TrueCar/gluestick-shared/pull/24

Also, we want to re-instate the errorHandler for any caught exceptions, rather than simply setting the status code on the response if there was a status code set in state. We will have a separate `renderErrorPage` action to handle cases where a user wants to render a different error page and status code.


saltycrane:
👍

PR #223, opened 2016-07-28
Check the version before npm updates

christinebrass:
Previously, the auto-upgrade functionality was taking place before the user was informed that their GlueStick version was out of date. Check the version first so the user applies the appropriate updates to their project.


JoeCortopassi:
👍

PR #225, opened 2016-07-28
Remove auto-upgrade code that is no longer needed

christinebrass:
This check prevents us from modifying the config file without receiving prompts by GlueStick that it is stale.  We have not changed the contents in the config file in a long while now, so I think it is safe to remove.


JoeCortopassi:
👍

PR #228, opened 2016-08-03
Use forever by default (but allow PM2 as an option)

christinebrass:


christinebrass:
Canceling this PR for now.

PR #233, opened 2016-08-09
Whitelist Greenkeeper branches

christinebrass:
Allow Greenkeeper to create branches against this project. https://greenkeeper.io


christinebrass:
I'm not sure whether this is truly needed. Greenkeeper seems to operate just fine without this change. Closing for now.

PR #235, opened 2016-08-10
Improve caching

toddw:
- Remove support for custom cacheKey which lets us return cache responses
  without calling `match`.
- Stream responses


christinebrass:
👍

PR #241, opened 2016-08-12
improve server only file support

toddw:
Before we were telling webpack to ignore certain files from the build but it was still throwing errors. This will return a noop function instead of just emitting it from the build.

This is experimental and possibly the wrong way to go about this…


christinebrass:
👍

PR #244, opened 2016-08-12
tell `serialize` it's JSON for INITIAL_STATE

toddw:
https://github.com/yahoo/serialize-javascript#optionsisjson

> This enables a hot-path that allows serialization to be over 3x
> faster.


christinebrass:
Nice 👍

PR #245, opened 2016-08-12
Switch to LRU cache

toddw:
Previously the instance's memory would explode if tons of dynamic pages are hit sync there was no limit to the number of items stored in a cache. That is solved by switching to a LRU cache with a max number of entries.

Also, it's surprising that the code previously worked by caching a ReadStream since they can only be read once. This caused the lru cache to break since the second read was returning an empty stream. This has been fixed by caching a buffer instead of the readstream. A readstream is still created and streamed out for the response.

Also, I noticed we were caching in dev mode so I don't cache unless we are in production mode now


src/lib/server/requestHandler.js
```diff
@@ -121,31 +120,27 @@ module.exports = async function (req, res) {
...
           }
 
-          // mark that we are done with the stream
-          responseStream.push(null);
-
           // If caching has been enabled for this route, cache response for
           // next time it is requested
-          if (currentRoute.cache) {
-            const cacheTTL = currentRoute.cacheTTL || DEFAULT_CACHE_TTL;
+          if (currentRoute.cache && process.env.NODE_ENV === "production") {
```
toddw, line 134: This line is worth calling out. We don't want to cache in dev mode

christinebrass:
👍

PR #250, opened 2016-08-15
Refactor client-side webpack configuration

christinebrass:
Separate out webpack config from the `start-client` command. With this, we also don't need the `build` command to be part of `start-client`. Also added tests for `detectEnvironmentVariables`.


toddw:
:+1:

PR #251, opened 2016-08-15
Add hash to js and css files

toddw:
For better CDN caching, we are now going to include hashes in the .js
and .css filenames the same way we do for other assets


christinebrass:
👍

PR #258, opened 2016-08-17
maxAge is milliseconds not seconds

toddw:
Even though `maxAge` sets the `Cache-Control` header which is in seconds,
the value expected from `maxAge` is in milliseconds. So need 3 more
zeros to get to 1 year.


christinebrass:
👍

PR #263, opened 2016-08-23
load the application code in an async tag

toddw:
Even though the script tags are at the end of our document, we can defer
the loading of the application code since the server did all of the
rendering for us.


christinebrass:
👍

PR #264, opened 2016-08-23
Rollback Async

toddw:
It seems as though the `async` attribute may have caused some issues in apps. Rolling back until it can be investigated closer.


christinebrass:
👍

PR #266, opened 2016-08-25
Make client build by default

christinebrass:
`gluestick start` should kick off a build unless it was told not to with the `-P` option


toddw:
:+1:

PR #268, opened 2016-08-26
Better way to handle process.env variables

toddw:


src/config/getWebpackClientConfig.js
```diff
@@ -74,6 +74,23 @@ export default function (appRoot, appConfigFilePath, isProduction) {
...
+              "babel-plugin-gluestick"
+            ],
+            presets: [
+              "react",
+              "es2015",
+              "stage-0"
+            ]
+          },
+          include: [
+            path.join(process.cwd(), "src/config/application.js"),
```
christinebrass: Can we use `appRoot` for `process.cwd()` here?

toddw: yeah totally! I didn't even know what that was

toddw: updated

christinebrass:
👍

PR #269, opened 2016-08-29
Performance tuning

toddw:
After running several load tests and analyzing the CPU profile we found
a few places that were showing up close to the top of the profiler.
Most of the time is spent in ReactDOM.renderToString but we are updating
the parts of the code that are part of GlueStick.

Move methods that were showing up in the JavaScript CPU profiler out of
the request handler that don't need to be run on each request. Add
memoizing to methods that can benefit from it.

Add another babel plugin that could improve performance.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,6 +6,7 @@
   ],
   "plugins": [
     "transform-decorators-legacy",
+    "closure-elimination"
   ],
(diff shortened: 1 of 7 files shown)


JoeCortopassi:
👍

PR #271, opened 2016-08-30
Perform build optimization on docker image

christinebrass:
Now that environment variables are interpreted at runtime (see #268), we no longer have the requirement to build upon server startup.  This is essentially an undo of #267 and https://github.com/TrueCar/gluestick/commit/9ee199d4d6e08e8a2c7b12fee4c0b1b62dc5bc37.


toddw:
:+1:

PR #275, opened 2016-08-31
Use req.url instead of req.path on the server

christinebrass:
According to react-router docs, we should be providing the full URL path from the original request, including the query string.

https://github.com/reactjs/react-router/blob/master/docs/guides/ServerRendering.md#server-rendering

This will fix an issue where `location` objects were dropping query params.


toddw:
:+1:

PR #277, opened 2016-09-02
move gzip to #1 priority

toddw:
IE11 has an issue with `deflate` so moving to `gzip` as the priority. IE11 literally wont render the page at all with deflate.


mikeheme:
A little late, but looks good 👍

PR #281, opened 2016-09-07
Stop using aggressive merging plugin

toddw:
When developers use code splitting to separate sections of an
application, the aggressive merging plugin will sometimes merge those
together. This can cause undesired behavior. If this is the desired
behavior then this plugin should be added on a per-app basis using the
webpack-additions.js file


christinebrass:
👍

PR #284, opened 2016-09-08
change devtool to `inline-source-map`

trungly:
This is to make it possible to debug client side code on Chrome again.

See issues:
https://github.com/webpack/webpack/issues/2846
https://github.com/webpack/webpack/issues/2145


toddw:
Thanks Trung, what are the speed implications of using `inline-source-map` vs `cheap-module-eval-sourcemap`? I'm looking at the chart here https://webpack.github.io/docs/configuration.html#devtool and I'm not seeing it listed. Although it is mentioned above the table. Have you tested it out to see the difference? These things can make a big impact when it comes to larger projects but I'm totally down to use the most useful one that doesn't slow things down too much.

toddw:
`cheap-module-eval-source-map` is supposed to show the original source (lines only), whatever lines only means… if it is not working, then that sounds like a bug we should look into. Are we using the latest version of WebPack?

trungly:
Gluestick is currently on Webpack 1.13.1.  I tried the latest (stable) version of Webpack (1.13.2), and the problem persists there.

Re: the other options for `devtool` and build performance, I tried nearly all of them.  None of the "eval" ones work (i.e. the debugger stops, but the source shows as "undefined").  Using `cheap-source-map` and `cheap-module-inline-source-map` also stop the debugger, but shows the compiled source... not very useful, imo.  The two I have found that work with the original ES6 source are `source-map` and `inline-source-map`.. There wasn't any difference between those, as far as I could tell.

I did some timings with the large consumer facing app we have here at work, and you're right, there is definitely a slow down.

Here are typical samples.

**cheap-module-eval-source-map**

```
webpack built 285a4bca30689f73a1b6 in 63316ms
webpack building...

Starting to optimize CSS...
webpack built 285a4bca30689f73a1b6 in 1345ms
```

**inline-source-map**

```
webpack built 5f268e16aa934f013bb8 in 75188ms
webpack building...

Starting to optimize CSS...
webpack built 5f268e16aa934f013bb8 in 1344ms
```

Again, `inline-source-map` and `source-map` were about the same.

So it's your call if this slow down is worth the ability to debug.  Perhaps we go with this for now and keep an eye out if Webpack will have the bug fixed.  I think it is beyond my ability to be able to figure out the bug in Webpack.

Of course, another option is make this option configurable per Gluestick app.  I explored this a little, but I'm not exactly clear if `getWebpackClientConfig.js` has access to the app's config.. or if not, if it's wise to read the config file from the file system at the point the function is called.

toddw:
:+1:

PR #285, opened 2016-09-08
use Optimize CSS Assets Webpack Plugin

toddw:
https://github.com/NMFR/optimize-css-assets-webpack-plugin

> ### Solves extract-text-webpack-plugin CSS duplication problem:
> 
> Since extract-text-webpack-plugin only bundles (merges) text chunks, if its used to bundle CSS, the bundle might have duplicate entries (chunks can be duplicate free but when merged, duplicate CSS can be created).


JoeCortopassi:
👍

PR #287, opened 2016-09-09
use `image-webpack-loader` to optimize images

toddw:
> Image loader module for webpack
> 
> > Minify PNG, JPEG, GIF and SVG images with imagemin
> 
> Issues with the output should be reported on the imagemin issue tracker.

https://github.com/tcoopman/image-webpack-loader


JoeCortopassi:
👍

PR #292, opened 2016-09-14
Less verbose test output

christinebrass:
Make tests less verbose by stubbing out logging calls rather than spying on them.

Old:
old output

New:
new output


toddw:
:+1:

PR #301, opened 2016-09-20
Stop including .server.js files in build

toddw:
.server.js files are not supposed to be included in the build but they were. It looks like the way I was using NormalModuleReplacementPlugin was different than the examples I could find. In the examples an actual file path was returned instead of a function. This seems to resolve that problem.


src/config/getWebpackClientConfig.js
```diff
@@ -114,7 +114,7 @@ export default function (appRoot, appConfigFilePath, isProduction) {
       }),
 
       // Make it so *.server.js files return empty function in client
```
christinebrass: `s/function/module/` ?

christinebrass:
👍

PR #305, opened 2016-10-05
add `gluestick run` command

toddw:
Just like Ruby on Rails has a `rails run` command that lets you run scripts within the context of your Rails application we needed the ability to do this for GlueStick applications. For now this just sets up webpack isomorphic tools and babel but this can be modified to include more things in the future


christinebrass:
👍

PR #306, opened 2016-10-05
add support for `config/init.server.js`

toddw:
Developers can add an optional file in src/config/init.server.js that includes any script initialization stuff. This is a good place to add something like `require("newrelic");`


christinebrass:
👍

PR #308, opened 2016-10-05
Load the vendor bundle earlier in the timeline

christinebrass:
There is no need to wait for DOMContentLoaded to be fired before starting to download the vendor bundle.


toddw:
:+1:

PR #309, opened 2016-10-06
Get rid of subresource links

christinebrass:
- They are deprecated
- They are usually downloaded with a low priority, which will not help with optimizing the download time anyway


toddw:
:+1:

PR #312, opened 2016-10-14
Pin Node version to 6.7

christinebrass:
Upgrading to 6.8.0 caused our builds to fail. Rather than wait for a
patch, lock down the version of Node to a version we know works for us.


toddw:
:+1:

PR #314, opened 2016-10-18
Refactor request handler

toddw:
The request handler is hard to test. This refactor is to push things in
a new direction where things are split up into multiple small
functions. These smaller functions are much easier to test and reason
about.
- [x] refactor to make testing easier
- [x] write tests
- [x] manual tests existing gluestick apps and new gluestick apps
  - [x] base app
  - [x] 404 using the no match constant
  - [x] 404 using status prop on route
  - [x] `gsBeforeRoute` success
    - [x] initial state is populated
  - [x] `gsBeforeRoute` failure
  - [x] single entry point
  - [x] multiple entry points
  - [x] cached response
  - [x] non cached response
  - [x] email render
  - [x] production mode


chefnobody:
👍

PR #316, opened 2016-10-21
filter html type from tests

andrewwatts:
a number of static file types have been excluded from mocha tests because they're not js modules, .html files should be filtered as well, eg: iframe content that is not react app.


toddw:
:+1:

PR #318, opened 2016-10-25
Fix test coverage commands

christinebrass:
The test coverage commands used a hard-coded path to the `babel-istanbul` executable.  They were also using the `babel-node` command which turned out to be unnecessary.


toddw:
:+1:

PR #319, opened 2016-10-26
Replace npm with yarn during docker image creation

christinebrass:
- Install yarn onto all GlueStick base images so apps can use yarn for faster installs
- Use yarn to install dependencies onto GlueStick base images
- Make yarn the default install tool for app docker images


toddw:
:+1:

PR #321, opened 2016-10-27
Switch to electrode-react-ssr-caching

toddw:
## breaking change

```
Anyone using component caching by providing a `componentCacheKey`
method will no longer get component caching after this change. You must
update and use the componentCacheConfig setting instead.
```

`react-dom-stream` has proven not to be super compatible with code that
needs to run through from top to bottom before being sent back to the
server. Namely, aphrodite css and server side rendering is not
compatible with a streaming renderer.

Honestly, we were never taking full advantage of streaming the page out
to the user before rendering the entire page because we use things like
Helmet to allow nested components to change content in the head tag.

We added `react-dom-stream` for the component caching.
`react-dom-stream` doesn't recommened using their component caching and
it has become incompatible with something we wanted to use. Luckily
electrode has a nice library that focuses soley on component caching
and not streaming.

This is a breaing change if you have been using the `componentCacheKey`
method to tap into component caching with `react-dom-stream`. The new
way to do component caching is by passing a componentCacheConfig
property in `src/config/application.server.js`. [See
electrode-react-ssr-caching docs](https://github.com/electrode-io/electrode-react-ssr-caching) for more information.

_Example src/config/application.server.js_

```
export default {
  componentCacheConfig: {
    components: {
      Home: {
        enable: true,
        strategy: "simple"
      }
    }
  }
};
```


christinebrass:
👍

PR #324, opened 2016-10-27
Updated README to recommend a specific major version of Node. Linked …

chefnobody:
…to known issue with Node v7.0.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ boilerplate code as new improvements are discovered.
 Is actively being worked on and supported by TrueCar.
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


christinebrass (changes requested):
Given that we now use features that were introduced in Node v6.0.0, we should probably also remove the blurb about supporting v4.

chefnobody:
Updated the bits about node v4. Thoughts?

christinebrass (approved):

christinebrass:
Looks good, thanks! 👍

PR #327, opened 2016-10-27
0.9.29

christinebrass:
Lock down dependencies better in gluestick-shared

https://github.com/TrueCar/gluestick-shared/pull/48


mishkinf:
👍

PR #328, opened 2016-10-28
Only use component caching in prod

toddw:
There is a bug when you try to enable component caching in development mode, so this makes it so that component caching is only done in production mode.


christinebrass:
👍

chefnobody:
👍 ☕️

PR #330, opened 2016-11-01
roll back to sync script loading

toddw:
We are getting error reports that suggest the script loading is
sometimes loading in the wrong order. Unable to reproduce the issue
but until the issue is resolved we are going to roll back to a more
synchronous approach for loading scripts.


christinebrass:
👍

PR #331, opened 2016-11-02
Add support for customizing asset settings

toddw:
Express static file server accepts a few arguments. This PR gives developers the ability to pass addition options to the static file middleware, as well as setting the path and build folder.

Express static server options: https://github.com/expressjs/serve-static

Example setting custom headers:
image


christinebrass:
👍

PR #334, opened 2016-11-02
Update dependencies to address vulnerabilities

christinebrass:
* Update express 4.13 -> 4.14.0
* Update mocha 2.5.3 -> 3.1.2
* Replace secure-handlebars with latest version of handlebars, as we're only using it to render the error page


toddw:
👍

PR #302, opened 2016-09-29
Pin babel to fix syntax errors

christinebrass:
Babel complains when running tests on a new GlueStick project:

```
› gluestick test
/Code/gluestick/node_modules/babel-template/lib/index.js:14
    if (error.stack) {
             ^
SyntaxError:
[GlueStick] ERROR: Coverage report generated, but with test errors.
^C
```

PR #337, opened 2016-11-04
Relax version requirements for babel packages

christinebrass:
We aren't getting accurate coverage reports for GlueStick with locked down versions of Babel. We have been affected by breaking babel-related changes in the past (see #302), but now that we are using yarn's lock files, I am somewhat more confident in this change than I would've been before.


christinebrass:
It's interesting that the `test-coverage` command seems to fail the build because dependencies are installed via `npm` and not `yarn`. Also, _only_ the `test-coverage` command fails. `npm test` is fine.

christinebrass:
Closing this since code coverage reports for GlueStick appear to be working as expected.

PR #338, opened 2016-11-08
fix assetPath bug

toddw:
When using `gluestick dockerize` the webpack-assets.json file is
generated which statically makes the assetPath stuck to what it was
when the build was created. This is a problem when you want to pass a
different environment variable to change assetPath at runtime.

Without a complete refactor of how the asset paths work, the current
solution is to keep track of what the assetPath was when `gluestick
build` was run. If, at runtime, a different assetPath is found, it will
replace the old one with the new one when calling `getAssetPathForFile`.


toddw:
temporarily closing until manual testing is completed

PR #342, opened 2016-11-09
Use LoadJS in place of script.js

christinebrass:
[LoadJS](https://github.com/muicss/loadjs) is very similar to script.js but it provides error handling via callbacks and is optimized for modern browsers. Also, it's very small at 654 bytes.

Unfortunately, loading the vendor and entrypoint bundles in parallel but as a series using the `async: false` flag still results in runtime error because the script will attempt to load the entrypoint bundle even if the vendor bundle fails to load.

To optimize, we can begin download of the vendor bundle before the `DOMContentLoaded` event fires. We can then wait for the vendor bundle to be ready before attempting to load the entrypoint bundle. We try to download the vendor bundle twice before throwing an error.


toddw:
:+1:

PR #346, opened 2016-11-11
Make render method more modular

toddw:
This gives the ability to pass a custom render method to gluestick for server side rendering. This is set from within `src/config/application.js`. This custom render method just needs to return HTML so that the server knows what to render.

The custom render method allows you to specify the body html as well as pass JSX to be included in the `head` tag.

This was mostly added to support Aphrodite which wraps the default render method and generates all of the CSS before returning. However, this gives developers more power over the HTML that is delivered to the client on server side requests.

**This is a power feature and not required to use GlueStick, if no custom `renderMethod` is passed in `src/config/application.js` it will continue functioning as before.**

Aphrodite example:
image


templates/new/src/config/.entry.js
```diff
@@ -28,6 +28,10 @@ export default class Entry extends Component {
 }
 
 Entry.start = function (getRoutes, getStore, match=originalMatch, history=browserHistory) {
+  // Allow developers to include code that will be executed before the app is
+  // set up in the browser.
+  require("./init.browser");
```
toddw, line 33: This is another power tool that lets developers execute code as soon as the app is mounted. Needed for Aphrodite:
image

src/lib/server/RequestHandler.js
```diff
@@ -159,15 +159,31 @@ export function prepareOutput(req, {Index, store, getRoutes, fileName}, renderPr
   // script tag that hooks up the client side react code.
   const currentState = store.getState();
 
+  let headContent, bodyContent;
+  const renderMethod = config.server && config.server.renderMethod;
+  if (renderMethod) {
+    try {
+      const renderOutput  = await Promise.resolve(renderMethod(reactRenderFunc, main));
```
toddw, line 166: Cool trick here… `Promise.resolve` is awesome. `await Promise.resolve(somethingHere)` when `somethingHere` is a promise, it will wait for that promise to resolve. If `somethingHere` is anything else, it will be ready immediately. This is a nice and clean way to accept a promise or a synchronous value

christinebrass:
👍

JoeCortopassi:
👍

PR #347, opened 2016-11-11
Remove vendor bundle loading optimization

christinebrass:
Vendor bundle downloads still seem to fail from time to time. We are going back to deferring the bundle download until after `DOMContentLoaded` to see if that makes any difference.


toddw:
:+1:

PR #349, opened 2016-11-11
Adding support to gluestick for 'gluestick-addons' which will enable …

mishkinf:
…building extensions to Gluestick written in es6 that will transpile with the webpack and babel configuration which your project is leveraging


christinebrass:
👍

PR #350, opened 2016-11-15
Make app code loading more resilient

toddw:
There seems to be failures mostly occuring on mobile devices. We used
to try reloading `vendor` once. This will attempt each javascript asset
up to 10 times before giving up.

I have also added a 100ms delay before each retry in case it was caused
by some kind of internet connection issue.

Instead of calling `window.__startGSApp` it will now start
automatically as soon as it is loaded. `_startGSApp` was added when
support for loading the app before vendor was loaded. Since we have
backed off from paralell loading due to loading errors we are seeing,
it is no longer necessary. The hope is removing that will also account
for less errors.

### DIFF
--- a/src/lib/buildWebpackEntries.js
+++ b/src/lib/buildWebpackEntries.js
@@ -100,7 +100,7 @@ export function getStore (httpClient) {
 }
 
 if (typeof window === "object") {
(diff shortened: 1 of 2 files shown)


christinebrass:
Looks good 👍

PR #352, opened 2016-11-16
Make asset path dynamic with environment variables

toddw:
Currently when creating a `build` for GlueStick, the asset path is hard coded into the webpack-assets.json file (which is what webpack-isomorphic-tools uses to get asset urls) and sometimes directly into assets themselves (like css files referencing a font file like you see from bootstrap-sass). If you know the asset url ahead of time then this isn't a problem, just make sure to specify the correct URL during the build step.

When you don't know the asset url during build time, this doesn't work out so well. In those situations we wanted to add support for dynamically setting the asset url even after a build has been created. This will let you use the same docker container in a QA env vs a production environment with only changing environment variables.

To support this we now create docker builds using the `ASSET_URL` environment variable set to `__GS_ASSET_URL__`. Then at runtime we set that URL based on the environment settings.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -17,6 +17,7 @@ const run = lazyMethodRequire("./commands/run");
 
 const updateLastVersionUsed = require("./lib/updateVersion");
 const getVersion = require("./lib/getVersion");
+const updateWebpackAssetPath = require("./lib/updateWebpackAssetPath");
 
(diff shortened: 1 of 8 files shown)


toddw:
This is not good enough yet… the contents of the CSS files get the path hard coded into it, need to come up with a plan for those

christinebrass:
👍

PR #353, opened 2016-11-16
CLI should use yarn when available

christinebrass:
This change allows the CLI to prefer `yarn` over `npm` when installing dependencies for a new GlueStick application and during the auto-upgrade process.


toddw:
:+1:

PR #356, opened 2016-11-23
support for static builds

toddw:


src/cli.js
```diff
@@ -98,8 +99,9 @@ commander
 commander
   .command("build")
   .description("create production asset build")
+  .option(...prepareStaticOption)
   .action(checkGluestickProject)
-  .action(() => build())
+  .action((options) => build(options.static))
```
toddw, line 104: just adding the option listed above so you can pass `--static` to the build command. The value from that will be true or false and passed to the `build` command in src/commands/build.js

src/commands/build.js
```diff
@@ -17,6 +18,11 @@ module.exports = function() {
     const statsJson = stats.toJson();
     fs.writeFileSync("webpack-bundle-stats.json", JSON.stringify(statsJson));
     fs.writeFileSync("build/asset-path.json", JSON.stringify({assetPath: getAssetPath()}));
+
+    if (staticBuild) {
+      generateStaticHTMLFile();
+    }
```
toddw, line 24: nothing too fancy going on here… during the build process, if `--static` was passed then we are going to have it build out the static html file.

src/lib/server/RequestHandler.js
```diff
@@ -128,7 +128,7 @@ export function enableComponentCaching (componentCacheConfig, isProduction, SSRC
   }
 }
 
-export async function prepareOutput(req, {Index, store, getRoutes, fileName}, renderProps, config, envVariables, getHead=_getHead, Entry=_Entry, _webpackIsomorphicTools=webpackIsomorphicTools) {
+export async function prepareOutput(req, {Index, store, getRoutes, fileName}, renderProps, config, envVariables, staticBuild=false, getHead=_getHead, Entry=_Entry, _webpackIsomorphicTools=webpackIsomorphicTools) {
```
toddw, line 131: All of the default variables you see on the end of this function were mostly added for dependency injection from a test. Please see the RequestHandler.test.js file to see how we take advantage of that. We try to isolate functions as much as possible so that they are easily testable.

src/lib/server/RequestHandler.js
```diff
@@ -150,18 +150,24 @@ export async function prepareOutput(req, {Index, store, getRoutes, fileName}, re
 
   // gather attributes that were included on the route in order to
   // determine whether to render as an e-mail or not
-  const currentRoute = getCurrentRoute(renderProps);
-  const routeAttrs = getEmailAttributes(currentRoute);
-  const isEmail = routeAttrs.email;
-  const reactRenderFunc = isEmail ? renderToStaticMarkup: renderToString;
+  let currentRoute, routeAttrs, isEmail, reactRenderFunc, currentState = {};
```
toddw, line 153: These variables only need to be set if we are doing actual server side rendering. In this case we just set them as empty variables and set them if we are _not_ doing a static build.

src/lib/server/RequestHandler.js
```diff
@@ -150,18 +150,24 @@ export async function prepareOutput(req, {Index, store, getRoutes, fileName}, re
...
 
   // grab the react generated body stuff. This includes the
   // script tag that hooks up the client side react code.
-  const currentState = store.getState();
 
   let headContent, bodyContent;
   const renderMethod = config.server && config.server.renderMethod;
-  if (renderMethod) {
+  if (staticBuild) {
+    bodyContent = "";
```
toddw, line 168: `bodyContent` is the html that will be inserted inside of `<div id="main"></div>`. GlueStick supports a default render method and custom render methods (in advanced situations). However, when we are doing a static build we really don't want to render anything so we just set the html to an empty string.

src/lib/server/generateStaticHTMLFile.js
```diff
@@ -0,0 +1,33 @@
+import path from "path";
+import { prepareOutput } from "./RequestHandler";
+import { getLogger } from "./logger";
+import WebpackIsomorphicTools from "webpack-isomorphic-tools";
+import webpackConfig from "../../config/webpack-isomorphic-tools-config";
+import detectEnvironmentVariables from "../detectEnvironmentVariables";
```
toddw, line 6: `detectEnvironmentVariables` is one of my favorite GlueStick features. When you build an app with GlueStick, the src/config/application.js file sometimes references environment variables for setting different config options.  Environment variables in node are accessed using `process.env.______`. Since this code is going to be run in the browser we needed a way to pass that data long to the browser. However, we don't want to pass all environment variables to the browser or else you would be leaking all kinds of stuff like AWS keys and other environment variables that need to be private. That is why you see a warning at the top of src/config/application.js that says any environment variables used in src/config/application.js will exposed in the browser.

We also didn't want to hard code the environment variables into the build because if you run the same build in multiple pods it is pretty common to change environment variables for different pod environments. So to deal with all of this we convert src/config/application.js into an AST and process it to find any references to an environment variable. Then we replace those environment variables with `window.__GS_ENV_VARS.____` and recompile the code. We also inject the actual values for those environment variables into the page.

Anyways, none of that needs to be updated but I figured it helps to understand what it is used for.

src/lib/server/generateStaticHTMLFile.js
```diff
@@ -0,0 +1,33 @@
...
+export default async function () {
+  const mockReq = {
+    headers: {
+      host: "localhost"
+    },
+    url: "/"
+  };
+
+  LOGGER.debug("about to run middleware");
+  global.webpackIsomorphicTools = new WebpackIsomorphicTools(webpackConfig)
```
toddw, line 19: the prepareOutput function relies on `webpackIsomorphicTools`. You can view more information about what that is here: https://github.com/halt-hammerzeit/webpack-isomorphic-tools but just know that the method needs to be run inside of the callback after configuring it.

src/lib/server/generateStaticHTMLFile.js
```diff
@@ -0,0 +1,33 @@
...
+    .development(false)
+    .server(process.cwd(), async () => {
+      try {
+        const envVars = detectEnvironmentVariables(path.join(process.cwd(), "src", "config", "application.js"));
+        const Index = require(path.join(process.cwd(), "Index.js")).default;
+        const output = await prepareOutput(mockReq, {Index}, {}, {}, envVars, true);
+        LOGGER.debug(output.responseString);
```
toddw, line 26: Currently I'm logging the responseString but really you want to write this to a file. The file would be `path.join(process.cwd(), "build", "index.html")` and you can do that easily with [fs.writeFileSync](https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options) we are okay using the synchronous version of `fs.writeFile` because this part is still under the command line interface. It is not used to process multiple incoming requests, so a blocking operation just makes the code easier to use without worrying about performance issues.

src/commands/build.js
```diff
@@ -2,12 +2,13 @@ import fs from "fs-extra";
 import path from "path";
 import webpack from "webpack";
 import getWebpackConfig from "../config/getWebpackClientConfig";
+import generateStaticHTMLFile from "../lib/server/generateStaticHTMLFile";
```
marcdel, line 5: @toddw `gluestick run` and `gluestick build` are currently broken on this branch. If I remove this import statement they work. I narrowed it down to lines [30-32](https://github.com/TrueCar/gluestick/blob/328845bc516f0df4ceec2a0bcd41a14401d8d7bf/src/lib/server/RequestHandler.js#L30) in RequestHandler.js. If I remove that block, `build` and `run` work (presumably because it's set to the same thing on line [134](https://github.com/TrueCar/gluestick/blob/328845bc516f0df4ceec2a0bcd41a14401d8d7bf/src/lib/server/RequestHandler.js#L134) if null?), but then `build --static` throws the same error that `build` and `run` were. I've tried a few things, but I think I'm stuck, at least as far as this piece is concerned.

```
gluestick build --static                                                                                              ⏎
/Users/marc/Code/static-gluestick/node_modules/react-transform-hmr/lib/index.js:51
    throw new Error('locals[0] does not appear to be a `module` object with Hot Module ' + 'replacement API enabled. You should disable react-transform-hmr in ' + 'production by using `env` section in Babel configuration. See the ' + 'example in README: https://github.com/gaearon/react-transform-hmr');
    ^

Error: locals[0] does not appear to be a `module` object with Hot Module replacement API enabled. You should disable react-transform-hmr in production by using `env` section in Babel configuration. See the example in README: https://github.com/gaearon/react-transform-hmr
    at proxyReactComponents (/Users/marc/Code/static-gluestick/node_modules/react-transform-hmr/lib/index.js:51:11)
    at Object.<anonymous> (/Users/marc/Code/static-gluestick/src/config/.entry.js:59:59)
    at Module._compile (module.js:556:32)
    at loader (/Users/marc/Code/marcdel/gluestick/node_modules/babel-core/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/marc/Code/marcdel/gluestick/node_modules/babel-core/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
```

toddw: You need to run it with `NODE_ENV=production` which will make it so that hot module replacement plugins are not applied

zamotany:
Already implemented.

PR #357, opened 2016-11-29
Fix server side stack trace

toddw:
Creating a new error object made it so we don't have the proper stack
trace when a server side rendering error occurs. This resolves that.


JoeCortopassi:
👍

saltycrane:
@toddw I ran into this yesterday also. Thank you for fixing. Why do you have the try/catch if you're only re-throwing the error?

toddw:
@saltycrane throwing an error because there is an `await` statement in the try block. Otherwise the promise would swallow the error.

PR #358, opened 2016-12-01
[WIP] move toward react-hot-loader

toddw:
We are still using react-hmr-transform which has been deprecated. Started looking into switching to `react-hot-loader`. As far as I can tell this should work but running into the same issues as this person: https://github.com/gaearon/react-hot-loader/issues/432


zamotany:
Already implemented.

PR #359, opened 2016-12-01
Fix dockerfile generation

toddw:
I recently updated the .Dockerfile that gets placed in application folders, however, I forgot to update the script that generates the .Dockerfile


saltycrane:
👍

PR #360, opened 2016-12-01
add support for webpack devtool override

toddw:
`inline-source-map` is slow but it was picked because once your code gets over a certain size it seems that webkit can't handle `cheap-module-eval-source-map`.

Until a better solution comes along this wil ladd the ability to at least override the `devtool` with `DEVTOOL=X` environment variable.

On my local machine I will be updating ~/.bash_profile to include
```
export DEVTOOL=eval
```
so my rebuilds are faster.

When I need the original code sourcemaps I'll manually override it. I rebuild more than I need to see the original source.

### Example
```
DEVTOOL=eval gluestick start
```


saltycrane:
I tried and it worked for me. This seems like a great idea. 👍

christinebrass:
👍

PR #367, opened 2016-12-06
Pin yarn version and update apt key

christinebrass:
Recent versions of yarn (v0.17) seem to have sporadic issues during the `yarn` install Docker step for installing the base set of dependencies for GlueStick apps. Pinning the yarn version to 0.16.1 seems to get rid of the problem.


toddw:
:+1:

PR #371, opened 2016-12-09
gluestick 0.10.15 (update gluestick-shared to 0.4.20 dispatch failure)

### LINKED ISSUE #354: promise middleware dispatches `*_FAILURE` for exceptions in app
When dispatching an action using the promiseMiddleware, sometimes an exception in the app code (due to a bug) will not show up as a stack trace in the console. Instead, the promiseMiddleware catches it and dispatches a `*_FAILURE` action. I'd expect the exception to be unhandled and that a `*_FAILURE` action only be dispatched if the promise in the original action rejects. I think this is considered "swallowing" exceptions.

I had seen this happen previously, but some recent discussion on Twitter prompted me to investigate:
https://twitter.com/dan_abramov/status/798838433368064000
https://twitter.com/dan_abramov/status/798856890813845504

Example gluestick project showing behavior: https://github.com/saltycrane/gs-examples/tree/master/dispatch-failure

Example snippets:

action:
```
export function myAction() {
  return {
    type: MY_ACTION,
    promise: new Promise((resolve) => {
      resolve("fulfilled");
    }),
  };
}
```

reducer:
```
export default (state=INITIAL_STATE, action) => {
  switch (action.type) {
    case MY_ACTION:
      throw "bug in my code";
      return "success";

    default:
      return state;
  }
};
```

On dispatching the result from `myAction()`, the "bug in my code" exception is not shown in the console, but the `MY_ACTION_FAILURE` typed action is dispatched instead.

An option for changing this behavior is to use the second argument of `.then()` instead of chaining the `.catch()` after the `.then()`. This was mentioned in the links above.

original code in promiseMiddleware (https://github.com/TrueCar/gluestick-shared/blob/c9d5533fd753d658c0965b760220c7133aa1da3c/src/lib/promiseMiddleware.js#L27-L35):
```
      return getPromise(client)
        .then(value => {
          next({...rest, value, type: SUCCESS});
          return value || true;
        })
        .catch(error => {
          next({...rest, error, type: FAILURE});
          return false;
        });
```

alternative code:
```
      return getPromise(client)
        .then(
          value => {
            next({...rest, value, type: SUCCESS});
            return value || true;
          },
          error => {
            next({...rest, error, type: FAILURE});
            return false;
          }
        );
```

maybe async+await could also be used?
```
      let value;
      try {
        value = await getPromise(client);
      } catch (error) {
        next({...rest, error, type: FAILURE});
        return false;
      }
      next({...rest, value, type: SUCCESS});
      return value || true;
```

saltycrane:
fixes https://github.com/TrueCar/gluestick/issues/354


toddw:
:+1:

PR #372, opened 2016-12-09
Lock down Travis yarn version to 0.16.1

christinebrass:
Releases are (silently) failing on the docker image creation step due to an issue with `yarn`. We locked down the version of yarn that is used within the docker images, but Travis is still using it's own version which seems to be causing the issue with the `postpublish` step. See https://api.travis-ci.org/jobs/182474194/log.txt?deansi=true as an example.


toddw:
👍

PR #373, opened 2016-12-12
Bump loadjs to 3.2.0

christinebrass:
* Use new `numRetries` option and remove old retry logic
* Set `crossOrigin` attribute to properly log script errors


toddw:
:+1:

christinebrass:
Leaving this as a wip PR for now, since we should make modifications to allow the loadjs `before` function to be configured on a per-app basis via the app config.

sethbattin:
@cybrass , I pulled the setting into consumer-frontend's PR 897.

christinebrass:
👍

PR #376, opened 2016-12-16
Address #124 and migrate server setting to the config file

### LINKED ISSUE #124: Allow webpack dev server to be externally accessed
Bind webpack-dev-server to 0.0.0.0 instead of localhost so that it can be accessed by machine IP.

I'm using another server (a VM in virtual network) as a frontend to React/node. Requests to to webpack-dev-server, including the request to /__webpack_hmr can't be proxied since it's bound to localhost and the VM is on a virtual network.

Specifically, I'm using nginx+php running in a VM as a front end that proxies server render requests to node, instead of accessing via localhost:8880.

See relevant issue here: https://github.com/webpack/webpack-dev-server/issues/147

sethbattin:
- change default host to 0.0.0.0 (thanks Todd)
- define the default host in the app config, and use that in start-client.js
resolves https://github.com/TrueCar/gluestick/issues/124


src/commands/start-client.js
```diff
@@ -17,7 +17,8 @@ const APP_CONFIG_PATH = path.join(APP_ROOT, "src", "config", "application.js");
 const ASSET_PATH = getAssetPath();
 
 const IS_PRODUCTION = process.env.NODE_ENV === "production";
-const PORT = 8888;
+
+const appConfig = require(APP_CONFIG_PATH).default;
```
toddw, line 21: This should fall back to `8888` if it isn't defined in the app config. That way old apps wont be broken

src/commands/start-client.js
```diff
@@ -41,7 +44,7 @@ module.exports = function () {
     // Proxy http requests from server to client in development mode
     app.use(proxy({
       changeOrigin: false,
-      target: "http://localhost:8880",
+      target: `${server.protocol}://${server.host}:${appConfig.assetPort}`,
```
toddw: same as above, appConfig.assetPort should fall back to `8880` so we don't break backwards compatibility on older generated apps

sethbattin:
@toddw another look, please?

src/commands/start-client.js
```diff
@@ -30,6 +31,9 @@ module.exports = function () {
 
   const compiler = webpack(getWebpackConfig(APP_ROOT, APP_CONFIG_PATH, IS_PRODUCTION));
 
+  const server = Object.assign({protocol: "http", host: "0.0.0.0", port: 8888}, appConfig.server);
+  const assetPort = appConfig.assetPort || 8880;
```
toddw, line 35: :+1:

toddw:
👍  Nice work @sethbattin! 🎉  This works great! Feel free to merge!

toddw:
closes https://github.com/TrueCar/gluestick/issues/124

PR #380, opened 2016-12-21
Static support

marcdel:
Hey @toddw / @cybrass,

I wasn't sure whether to PR against `develop` or `toddw/static-support`, but the later was a bit noisy since the branch is out of date. Happy to make another one there instead if you'd rather though.

The feature itself is now working (I was able to create/build a [test app](http://gluestick-static-test.s3-website-us-east-1.amazonaws.com/) with my branch and upload the `build` folder to S3 and it works great), and normal start/build commands seem to be working again as well.

I could use some guidance on the tests. I'm a bit stumped writing the tests for generateStaticHTMLFile (see the PR [here](https://github.com/marcdel/gluestick/pull/1) for weirdo webpack issues). Also, some of the tests I wrote for RequestHandler.prepareOutput may not be necessary and there may be some that I missed. Open to any suggestions.


src/lib/server/RequestHandler.js
```diff
@@ -27,7 +26,7 @@ let _Entry;
  * because the tests move around in folders and it wont exist when this file
  * is initialized but instead when the method is run.
  */
-if (process.env.NODE_ENV !== "test") {
+if (process.env.NODE_ENV === "production") {
```
marcdel, line 30: This is potentially controversial since I'm not sure the original reason it was added. As far as I can tell now that `generateStaticHTMLFile` is being imported in build.js this will fail unless NODE_ENV is production anyway.

christinebrass:
Looks like the latest build of this contains issues with the `-Z` option. Somehow `config.server` is `undefined` in this case only, causing this line to throw: https://github.com/marcdel/gluestick/blob/0a7356e70a3e6a75cb4c5773cf5e19ae96ae8bb3/src/lib/server/getHead.js#L37

christinebrass:
Looks good! We'll just want to base this PR on top of `v1` now and test any other changes that might be included. 👍

marcdel:
Closing in favor of https://github.com/TrueCar/gluestick/pull/539

PR #382, opened 2016-12-29
Move application server details to application.server.js

toddw:
Config data placed in _src/config/application.js_ will be bundled into
the browser bundle. It seems like this configuration data is server
specific and so I'm placing it in _src/config/application.server.js_ so
that it is not included in the browser bundle. We already support
_application.server.js_ for other things like overriding the server
side rendering method.


christinebrass:
Looks good, thanks! 👍

PR #383, opened 2016-12-29
Fix HMR in IE11

toddw:
This fixes Hot Module Replacement in browsers that don't support
EventSource natively by including an EventSource polyfill before
webpack-hot-middleware in non production environments


christinebrass:
👍

Issue #345, opened 2016-11-10
Cover reports no longer being generated in gluestick (resolve with Jest)


christinebrass:
The generation of coverage reports broke, likely after the upgrade to Mocha 3 in #334.

christinebrass:
Re-opening this issue since the change was reverted in #402.

toddw:
This is related and might be resolved by the work done in #406

PR #384, opened 2016-12-30
use babel-node for coverage

saltycrane:
Fix https://github.com/TrueCar/gluestick/issues/345

This made the coverage report work for me. See if this works for others also.


christinebrass:
This works for me now! Thanks for fixing! 🎉

PR #385, opened 2016-12-30
improve shared dependency management

toddw:
## Problem
The GlueStick and it's generated apps share a lot of dependencies. Some of these dependencies are shared by the CLI and some (most) are shared by the server side rendering component of GlueStick. This is because GlueStick keeps the server side rendering logic outside of the generated apps. The existing solution is to match shared dependency versions so that the generated apps share the same version as GlueStick. This is important for many components like React, if the server side rendering is done with one version of React and the browser uses another, the resulting HTML can likely be very different.

The sad thing about the current solution is that apps are dependent on GlueStick to update dependencies that it shares in order to update their apps. While we try to keep up to date, we realize that it is annoying to have to wait for GlueStick to update just to update something like React.

## Solution
This update utilizes [require-hacker](https://github.com/halt-hammerzeit/require-hacker) to add to the locations require tries to resolve from for packages that are known to be shared.

## Notes

- [x] ~This was able to get rid of most shared dependencies but there still are a few babel-preset babel related things to keep in sync. We need to determine how to handle that~
already handled with previous tooling
- [x] ~determine how to check for "required" dependencies. For example, the electrode-react-ssr-cache module, it _must_ exist for things to work, how do we want to check and verify that it exists?~
already handled with previous tooling
- [x] determine how to specify minimum version numbers.
- [x] update tests to validate new logic
- [x] manually test a couple existing apps to make sure things still work

For the first to-do list item, our current tooling might already solve this for us because it already tries to match versions on shared packages.

For the last two to-do items… If it exists in templates/new/package.json then it _must_ exist in the app. Likewise, the version in the app should be >= to the version in templates/new/package.json for things to work. We can use our current auto upgrade tooling to make this as painless as possible.


src/lib/testHelperShared.js
```diff
@@ -1,3 +1,4 @@
+require("./projectRequireHack");
```
threehams, line 1: projectRequireHack requires runThroughBabel which requires babel-polyfill. The next line also requires babel-polyfill. Isn't this going to error out?

toddw: it's not error'ing out

christinebrass:
This looks good to me 👍

PR #386, opened 2016-12-30
[WIP] Remove Radium Integration

toddw:
Radium has become an app level concern and many apps are using things like StyledComponents or Aphrodite. Radium should still be possible but not forced into every app.

Hopefully we don't just remove Radium but make it "pluggable" or easily "swappable". Maybe we want to introduce style strategies and we can provide "Radium" and "Aphrodite" out of the box while supporting custom strategies as well.


zamotany:
Radium was already removed and extracted to plugin.

PR #388, opened 2017-01-03
vagrant-based gluestick environment

sethbattin:


sethbattin:
@JoeCortopassi I can give you a demo of how I am using it for our internal projects, too

PR #389, opened 2017-01-03
Update yarn install for travis-ci

christinebrass:
The yarn docs now contain install instructions for installing specific versions of yarn for Travis CI. This removes the curl-based approach in favor of their recommended method:

https://yarnpkg.com/en/docs/install-ci#travis-tab

We are also updating the $PATH variable so that Travis CI will  use the version of yarn that is installed by us over the default installed version of yarn.


toddw:
:+1:

Issue #390, opened 2017-01-03
Legacy projects fail to build if src/config/application.server.js is missing


toddw:
https://github.com/TrueCar/gluestick/blob/f52e1748ab8b4ce3ab0a4365371e20927c554ccd/src/commands/start-client.js#L33

should use `loadServerConfig` from
https://github.com/TrueCar/gluestick/blob/1370323db0ce3a78329de378fbf4568dade2e428/src/lib/server/loadServerConfig.js

toddw:
should be solved by https://github.com/TrueCar/gluestick/pull/391

PR #391, opened 2017-01-04
Resolve issues in legacy apps when _src/config/application.server.js_ doesn't exist

mishkinf:
…l for when there is no application server config file


toddw (approved):
nice work!

toddw:
:+1:

toddw:
solves #390

PR #393, opened 2017-01-06
[WIP] adding in postcss support to gluestick

mishkinf:
Initially I also added in support to autoprefixer directly into gluestick. instead i changed to make gluestick configurable and autoprefixer can be included as needed by the client application


christinebrass:
@mishkinf is this PR still needed? If so, can you resolve conflicts and fix tests?

zamotany:
Already implemented

PR #395, opened 2017-01-10
make dnsmasq more portable

toddw:
This PR takes and slightly modifies https://github.com/TrueCar/gluestick/pull/298 by @regiswilson

> We should not override the container's resolv configuration (which
> might have important configuration information in it). Instead, we add the
> nameserver for localhost to be inserted above the container's name servers, which
> dnsmasq will then use to resolve addresses.

> Also, allow other files to be placed in /etc/dnsmasq.d so that
> subsequent docker configuration layers can be added as needed.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -71,7 +71,6 @@ RUN apt-get update && \
 
 RUN apt-get update && apt-get install dnsmasq -y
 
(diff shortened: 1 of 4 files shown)


christinebrass:
👍

PR #396, opened 2017-01-10
Move babel-traverse to app to fix webpack builds

christinebrass:
`gluestick build` commands fail for some apps, possibly due to the way babel is resolving plugins. To fix this, we move `babel-traverse` to the app so that both `babel-core` and `babel-traverse` are using the same babel instance.


toddw:
:+1:

PR #397, opened 2017-01-11
Ignore any null CLI logger options

christinebrass:
By default, the CLI sets the logger level to `null` if one wasn't specified: https://github.com/TrueCar/gluestick/blob/develop/src/cli.js#L89.

We should make sure that `null` options don't override valid options that were set via the app config or some other means.


toddw:
:+1:

PR #402, opened 2017-01-11
Replace babel-node with node 

christinebrass:
Using babel-node in the test command causes an issue with requiring multiple instances of babel-polyfill. This reverts that for now until a proper refactor of the test command can be performed.


saltycrane:
👍

toddw:
:+1:

PR #404, opened 2017-01-12
remove ReactTestUtils global in test helper

toddw:
No one probably uses it anymore because enzyme makes things way easier.
You can include it as a project level dependency if you want to use it.
It's no longer in it's own module but part of `react-dom` so if we
wanted to keep the global we would have to force people to update to a
newer version of React. That's okay but again, no one uses it so let's
just get rid of it


cybrass:
👍

christinebrass:
👍

PR #411, opened 2017-01-13
Upgrade react/electrod-cache/enzyme

toddw:
In order to upgrade React we also had to update enzyme, and
electrode-react-ssr-caching.

Updating these libraries exposed a bug in the test script that wasn't
using a mocked version of `loadNewProjectPackage` so that was fixed as
well.


toddw:
New apps still have a number of warnings that need to be resolved before we merge this:

```
webpack:///./~/fbjs/lib/warning.js?:36` Warning: Unknown props `radiumConfig`, `_radiumDidResolveStyles` on <div> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by StyleRoot)
    in StyleRoot (created by Entry)
    in Entry
printWarning @ webpack:///./~/fbjs/lib/warning.js?:36
```


```
Warning: You are manually calling a React.PropTypes validation function for the `path` prop on `Route`. This is deprecated and will not work in production with the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.
```

```
webpack:///./~/fbjs/lib/warning.js?:36 Warning: You are manually calling a React.PropTypes validation function for the `children` prop on `Provider`. This is deprecated and will not work in production with the next major version. You may be seeing this warning due to a third-party PropTypes library. See https://fb.me/react-warning-dont-call-proptypes for details.
printWarning @ webpack:///./~/fbjs/lib/warning.js?:36
```

toddw:
Need to update `gluestick-shared`

toddw:
The errors mentioned previously have ben resolved

christinebrass:
👍

PR #414, opened 2017-01-13
use replace-in-file instead of shelljs.sed

toddw:
Using `replaceInFile` instead of shelljs.sed because shelljs.sed reads files into an array, then runs replace over each line before writing back to the file. This was a problem when trying to read binary files like images.  `replace-in-file` has better code that doesn't try to split the file into an array, then join it back and it wont change the file if nothing is changed.


christinebrass:
👍

PR #415, opened 2017-01-13
Add react-addons-test-utils as a dev dependency

christinebrass:
The `gluestick test` command fails for new projects because the dependency is missing from them. This adds it for any projects and should cause an auto-upgrade for already-existing apps that are missing the dependency.


toddw:
:+1:

PR #425, opened 2017-01-18
proxy `npm test` to gluestick rather than an unconfigured error message

sethbattin:
The default npm test message is unhelpful (and incorrect).  Hopefully this increases ease of use for green users.


christinebrass:
Looks good, thanks! 👍

PR #433, opened 2017-01-20
Webpack config exposal

michalchudziak:
I have mixed feeling about replacing whole `webpack-additions.js` with a regular webpack config file, that's why I've exposed another hook for `additionalWebpackConfig`. 

I like the way current hooks works and if we decide to drop it completely I will require a lot of additional changes in the way GlueStick currently works, and I'm afraid it might break developer experience a bit.

Let me know what do you think about this solution. Do you like it, or should we remove it completely?

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,8 @@
     "url-loader": "0.5.7",
     "webpack-dev-middleware": "1.8.1",
     "webpack-hot-middleware": "2.12.0",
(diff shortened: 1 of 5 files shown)


toddw (approved):
I think this could work. Let's start with this and if we need more flexibility in the future then we can update it. I like how it maintains the backwards compatibility with previously generated apps.

👍

Issue #381, opened 2016-12-21
Use production ready `devtool` in webpack when NODE_ENV==="production"


jer-k:
And minified reports are very hard to discern. A few examples https://rollbar.com/TrueCar/consumer-frontend/items/215/ and https://rollbar.com/TrueCar/vipr/items/1661/

A thought is that the SourceMaps for the projects need to be included in production. 

A potential location for the issue is https://github.com/TrueCar/gluestick/blob/develop/src/config/getWebpackClientConfig.js#L72

And more information about Rollbar + SourceMaps https://rollbar.com/docs/source-maps/

michalchudziak:
@toddw @cybrass can you confirm that we want to implement it this way? Do you have some more thoughts on this case?

toddw:
Yeah, I think _at least for now_ it would be reasonable to make the default production mode `devtool` to be equal to `source-map` as seen in this table: https://webpack.github.io/docs/configuration.html#devtool

If people want to customize it further they can probably take advantage of #407 but using `source-map` seems like a reasonable default

PR #434, opened 2017-01-20
Set source-map as default production devtool

michalchudziak:
Solves #381 

Now `source-map` is default webpack setting for production builds.


toddw (approved):
👍

Issue #140, opened 2016-05-06
Version update script


christinebrass:
Create a version update script that:

1) Updates our npm package version
2) Updates the version in our Dockerfile: https://github.com/TrueCar/gluestick/blob/develop/templates/new/src/config/.Dockerfile. This is already taken care of by the `npm run prepublish` script, but when releasing via GitHub this doesn't get committed in time.

michalchudziak:
About the second part of this task - I'm not really familiar with GitHub releases. Can you tell us a bit more about how are you handling releases in this project? Do we have some hook where we can attach it (just like `npm run prepublish`) or should we think about some different ways of solving this issue?

toddw:
I can fill in some context as to what we do with GitHub releases but I'm going to need to rely on @cybrass to fill in the details for this specific issue:

We have a hook in Travis CI that publishes GlueStick to npm and our docker image to https://hub.docker.com/ when you create a new release. (https://travis-ci.org/TrueCar/gluestick)

Releases can be seen here: https://github.com/TrueCar/gluestick/releases

christinebrass:
Yeah, the way we have it setup now would work best if we were running `npm publish` locally. If we were, this would be our process:

1. Update the version in package.json 
2. Run `npm publish`. This would invoke our prepublish and postpublish scripts for prepping and uploading our GlueStick Docker image
3. Commit the changed .Dockerfile file
4. Tag the set of files and push the tag to GitHub

Releasing with GitHub releases is a little different, in that it tags the set of files for you when you "publish" the release. Here's the process we currently follow with GitHub releases:

1. Update the version in both the package.json file and the .Dockerfile file
2. Commit and push those changes to GitHub
3. Use the GitHub Releases section to draft and publish a new release. This is where we enter the git tag we want to use for the release

From there, our [Travis CI hook](https://github.com/TrueCar/gluestick/blob/develop/.travis.yml#L37-L44) does the `npm publish` step, which invokes the same prepublish and postpublish scripts for prepping and uploading the GlueStick Docker image to https://hub.docker.com/r/truecar/gluestick

I originally created this issue to solve Step 1 of this latter process, since in the beginning we were accidentally forgetting to change the version change to the .Dockerfile file after changing the version in package.json. I was thinking it would be nice to run a single script to update both the package.json file and the .Dockerfile at the same time. It's sorta trivial, but for someone new to the project it would be less of a hassle to remember to update both files. I'm totally open to suggestions to make the process better if you have any.

[...]

PR #435, opened 2017-01-20
Script to update package version

lukewalczak:
Solves #140

Added script to package.json to update GlueStick version.

Example usage: 
`npm run update-version patch`


toddw (approved):
I didn't realize you could do things like _postX_. I assume that will run automatically after running whatever X is? Very cool!

👍

Issue #355, opened 2016-11-22
Expose express app object to application


toddw:
Sometimes apps need deeper access to the express application. For instance setting `trust proxy` or other express specific things. It would be nice if there was something that could go in `src/config/application.server.js` which maybe looked like this:
```
function handleIt (app) {
   app.enable("trust proxy");
}

export {
  expressAppHook: handleIt
}
```

Not sure what we want to call it `expressAppHook` is kind of long, `appHook` isn't very descriptive but open for discussion. This should be easy to add, we would just import `expressAppHook` from the application.server config file after setting up the express app and pass the app as an argument during setup.

lukewalczak:
How should this hook look like ? I assume that `trust proxy` is not only thing we want to customize.

toddw:
It really just needs to expose the express app to a function. From there they should be able to do whatever they want with it. This will allow developers to specify settings, add route handlers etc…

toddw:
For more context I would say it should work like this:
1. Check`expressAppHook` is defined (open to other names… `expressAppHook` works but I wouldn't say I love it)
2a. If not defined… don't do anything
2b. If defined then simply call it from above maybe this line here: https://github.com/TrueCar/gluestick/blob/develop/src/lib/server/index.js#L39
and it would look something like `expressAppHook(app)`

Another example of a hook:
```
// src/config/application.server.js
import appHook from "./setupApi";

export {
  expressAppHook: setupApi
}
```

```
// src/config/setupApi.js
export default function (app) {
  app.get("/api/items", (req, res) => {
    // return items
  });

  app.post("/app/items", (req, res) => {
    // add item
  });
}
```

PR #436, opened 2017-01-20
Added expressConfigurator function to expose express app

lukewalczak:
Solves #355 

Added expressConfigurator to expose the express app to a client function.
Propose `expressConfigurator` name, it's more descriptive for me.


toddw (approved):
👍 I like the proposed name!

Issue #406, opened 2017-01-12
Refactor Test Suite inside GlueStick project


toddw:
There are a few issues with the current test suite:
1. Debugging tests is currently broken. We would like to use https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js
2. We no longer use Karma, all Karma stuff should be removed.
3. It feels as though the additional `spawn` might be unnecessary although it was said to have been added for speed.
4. Decouple coverage reports from running every time a test is re-run.

threehams:
The spawn with --watch was necessary at the time buuuuut... is definitely not the best solution, especially since --watch is being deprecated.

Any solution that allows incremental compilation would work here. Mocha 2 has a complicated and undocumented programmatic method for running tests. If Mocha 3 has improved that, and we can use chokidar to watch files and programmatically re-run tests with incremental compilation, that'd be the best of everything.

If we just restart mocha each time, we're looking at 12-15 seconds on every file change before tests even run. This is what caused us to add an alternative test runner into the app to get back to 10ms.

We do have to be careful not to run tests and the server in the same process, or wacky stuff happens. Been there before.

toddw:
It would be nice to be able to take advantage of mocha's many options straight from the CLI. For example, the --grep option for selecting a set of tests to run.

threehams:
I agree, but Mocha's forcing our hand a bit on that one by getting rid of `--watch`. I'll try to find the discussion a little later if you need it for reference.

toddw:
Maybe then, it would be worth considering Jest instead?

toddw:
It may not be so painful to convert to Jest with things like https://github.com/skovhus/jest-codemods

When we started GlueStick, there was a lot of talk about Jest looking like it was basically abandonware… The Jest I'm seeing today looks incredibly different than before.

My favorite items on the features list
1. Fast interactive mode with --watch (yay!)
2. Jest works with any compile-to-JS language and integrates seamlessly with Babel.
3. Create coverage reports with --coverage. No additional setup or libraries needed! (maybe we wont have istanbul problems anymore)
4. Run tests in parallel processes to minimize test runtime.

Some of these might overlap with Mocha but those _might_ be a lot cleaner/simpler/faster?

threehams:
Relevant discussion:
https://github.com/mochajs/mocha/issues/1780#issuecomment-232753427

threehams:
[...]

Issue #333, opened 2016-11-02
Address security vulnerabilities found with nsp


christinebrass:
Some of our dependencies contain security vulnerabilities. Let's do what we can to address these.

```› nsp check
(+) 6 vulnerabilities found
┌───────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                                                                                                                                                                               │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Name          │ minimatch                                                                                                                                                                                                          │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Installed     │ 2.0.10                                                                                                                                                                                                             │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Vulnerable    │ <=3.0.1                                                                                                                                                                                                            │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Patched       │ >=3.0.2                                                                                                                                                                                                            │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Path          │ gluestick@0.10.2 > babel-istanbul@0.8.0 > fileset@0.2.1 > minimatch@2.0.10                                                                                                                                         │
├───────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/118                                                                                                                                                                             │
└───────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                                                                                                                                                                               │
[...]

Issue #322, opened 2016-10-27
gluestick app won't start when run with Node v7.0.0


chefnobody:
Using:

```
Aarons-MacBook-Pro:gluestick aconnolly$ node -v
v7.0.0
Aarons-MacBook-Pro:gluestick aconnolly$ npm -v
3.10.8
```

And running `npm install` from a freshly forked `gluestick` repo results in the following errors:

```
Aarons-MacBook-Pro:gluestick aconnolly$ npm install

> v8-debug@0.7.7 preinstall /Users/aconnolly/workspace/gluestick/node_modules/.staging/v8-debug-c1f18b53
> node -e 'process.exit(0)'


> v8-debug@0.7.7 install /Users/aconnolly/workspace/gluestick/node_modules/v8-debug
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! Tried to download(404): https://node-inspector.s3.amazonaws.com/debug/v0.7.7/node-v51-darwin-x64.tar.gz 
node-pre-gyp ERR! Pre-built binaries not found for v8-debug@0.7.7 and node@7.0.0 (node-v51 ABI) (falling back to source compile with node-gyp) 
  CXX(target) Release/obj.target/debug/src/debug.o
../src/debug.cc:38:27: error: too few arguments to function call, expected at least 2, have 1
        v8::Debug::Call(fn);
        ~~~~~~~~~~~~~~~   ^
/Users/aconnolly/.node-gyp/7.0.0/include/node/v8-debug.h:198:3: note: 'Call' declared here
  static MaybeLocal<Value> Call(Local<Context> context,
  ^
../src/debug.cc:46:61: error: too few arguments to function call, single argument 'isolate' was not specified
        Isolate* debug_isolate = v8::Debug::GetDebugContext()->GetIsolate();
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
/Users/aconnolly/.node-gyp/7.0.0/include/node/v8-debug.h:248:3: note: 'GetDebugContext' declared here
  static Local<Context> GetDebugContext(Isolate* isolate);
  ^
../src/debug.cc:61:67: error: too few arguments to function call, single argument 'isolate' was not specified
        Local<Context> debug_context = v8::Debug::GetDebugContext();
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
[...]

PR #437, opened 2017-01-20
Refactor test suite to Jest (+watch +coverage)

ferrannp:
Related to #406 

- All tests from `test` folder are now using Jest and they only depend on it (no `sinon`, no `expect`). There is a configuration in `package.json` to tell Jest only run tests inside `test` folder
- `--watch` mode is working (see gif below)
- Running `npm run test-coverage` will create html page into `coverage` folder (so it also solves #345)
- Debugging with Chrome, just need to follow these steps: https://facebook.github.io/jest/docs/troubleshooting.html#tests-are-failing-and-you-don-t-know-why 

This is a pull request to get familiar also with Jest and play with it (specially with the `--watch` mode). If we are happy with it, we could also introduce Jest to the generated app with gluestick so we can get rid of `istambul` (solving #333), `mocha`, `karma`, `sinon` and `expect` (maybe even more if we needed special `babel` deps for it).

## Watch mode

watch

## Debugging

screen shot 2017-01-20 at 17 53 11

screen shot 2017-01-20 at 17 53 16


sethbattin:
Followon from #322 : my first thought is this:  https://youtrack.jetbrains.com/issue/WEB-14979 <- jest does not have direct webstorm support yet.  :|

But that's really just ide-sugar anyway.  As long as there is a cli entry point and file saving triggers watch mode, then we're good regardless of explicit support.  And if debugging via devtools works, then that solves the other use case.

I will try this one out in earnest later in the day.

zamotany:
>Debugging with Chrome, just need to follow these steps: https://facebook.github.io/jest/docs/troubleshooting.html#tests-are-failing-and-you-don-t-know-why

They use `node-inspector` but we are removing it in favour of `node --inspect` (#322).

zamotany:
https://youtu.be/1cfA87h_1hI
on Ubuntu 16.04

Notice that `debugger;` statement in test **doesn't** trigger break.

toddw:
This looks really cool!

Looks like there are some issues in the build: https://travis-ci.org/TrueCar/gluestick/builds/193784640#L387

toddw:
The tests are broken specifically when `--watch` is passed so it is likely something that just needs to be updated in our tests

ferrannp:
All test seem to `PASS` but for some reason the build gets stuck (not exiting with error or success). I am not sure if those console.error(s) or some throwings that are happening in some tests are the cause. I need to investigate/try more on this.

---
Edit 1:

Ok, I found that the test that makes Travis to fail is `new.test.js`, that tests promts the user to a question `Do you wish to continue?` and this keeps Jest to be hanging. I need to investigate a solution for this.

The other thing that concerns me is that all these tests are actually hitting the disk and creating/removing files. Probably, we want unit tests to mock this behaviour. See the exact example with `fs` https://facebook.github.io/jest/docs/manual-mocks.html

---
Edit 2:

Ok I have the build passing: https://travis-ci.org/TrueCar/gluestick/builds/194489121

The problem is that I had to run it manually since it looks like that PRs to `TrueCar:v1` are not triggering builds in Travis. Can you check this @toddw? I guess it is just to change `travis.yml` in `develop`.

toddw:
I updated _.travis.yml_ so it should now include the `v1` branch

toddw:
> The other thing that concerns me is that all these tests are actually hitting the disk and creating/removing files. Probably, we want unit tests to mock this behaviour.

Yeah, since a major part of the CLI is creating files, it does write and read from disk during tests. I'm sure we could mock this but it might get tricky to validate the CLI is actually doing what it should. We are totally open to any suggestions here

ferrannp:
So build is green :). About hitting the disk, I think that for now is fine. Maybe in the future we want to try something but it does not feel as a bottleneck right now. I think this PR is ready as it is @toddw.

toddw:
works great! Thanks @ferrannp 

I was not able to get Chrome inspector to work yet but that's a different issue.
Related (https://github.com/TrueCar/gluestick/issues/406#issuecomment-274439629)

ferrannp:
@toddw you should be able when https://github.com/TrueCar/gluestick/pull/445 is merged.

PR #445, opened 2017-01-23
Use built-in  node inspector instead of node-inspector

zamotany:
Fixes: #322

Review and merge this after #437 is merged.


toddw:
Running `npm run debug` results in 24 tests failing… They all seem to fail with the same error message
> only one instance of babel-polyfill is allowed

zamotany:
@toddw currently you can debug only one test at once:
```
nom run debug <path/to/test>
```
I wrote quick note about it in `README.md`.

This is a known issue when debugging all tests at once, but currently I ran out of ideas how to fix it. Nevertheless limitation to debug only one test at once is somewhat logical and improves performance, since there is no need to wait for other tests to finish.

I suggest to leave this limitation for now and see in practice if option to debug all tests at once is actually needed and then take care of it.

toddw:
:+1:

toddw:
Made a couple updates here: https://github.com/TrueCar/gluestick/pull/453

PR #450, opened 2017-01-23
`gluestick new` success message

ferrannp:
- The test needs to be reviewed after #406 
- Of course the message can be edited if we feel we want to add/remove something


src/commands/new.js
```diff
@@ -29,8 +29,8 @@ function copyTo (destination) {
 module.exports = function (projectName) {
   const currentlyInProjectFolder = isGluestickProject();
 
-  // No project name, or ran from inside an existing project install in current directory if approved
-  if (!projectName || currentlyInProjectFolder) {
+  // Ran from inside an existing project, install in current directory if approved
+  if (currentlyInProjectFolder) {
```
ferrannp, line 33: I checked that if you do not pass `projectName` we actually throw an error so this condition could be simplified.

src/commands/new.js
```diff
@@ -54,6 +55,13 @@ module.exports = function (projectName) {
...
   // Project name set, install in current working directory
   copyTo(path.join(process.cwd(), projectName));
+  _printInstructions(projectName);
   return true;
 };
 
+function _printInstructions(projectName) {
+  logger.info(`${highlight("New GlueStick project created")} at ${filename(process.cwd())}`);
+  logger.info("To run your app and start developing");
+  logger.info(`    cd ${projectName}`);
```
toddw, line 65: should we add a line after this that says `     gluestick start`, before they can get anything in their browser?

toddw (changes requested):

src/commands/new.js
ferrannp: Yup, done :)

Issue #410, opened 2017-01-12
Allow httpClient to be accessed within onEnter hooks


toddw:
Currently, `httpClient` is only accessible via the promiseMiddleware. In an action creator, if you pass a function to the `promise` key it will get the `httpClient` passed to it. The advantage of using this instead of plain axios, is that this will be the configured axios client. That way if you pass a value to `httpTransforms` in src/config/application.js those will be applied to the httpClient. Also, on server side requests, relative URLs will be automatically converted to full URLs so that they will work in server side requests. Additionally, on server side requests, cookies are also grouped together for you via https://github.com/TrueCar/gluestick-shared/blob/master/src/lib/getHttpClient.js

Really we want to be able to pass the result of `getHttpClient` to an `onEnter` hook but in order to do so we need to be able to pass along `options={}, req, res,`. `req`, and `res` are references to express request and response objects generated by a server side request. They are not required when building an http client in the browser so they are optional arguments but they should be passed when being run on the server.

michalchudziak:
Are you talking about *react-router's* `onEnter` hook in generated project or anything else? Would be nice if we could clarify this one :)

toddw:
Yeah sorry that is exactly what I'm talking about. _react-router's_ `onEnter` hook.

toddw:
resolved by https://github.com/TrueCar/gluestick/pull/455

PR #455, opened 2017-01-24
#410 onEnter hook added

olga32123:
Example of usage :

```
/* @flow */
import React from "react";
import { Route, IndexRoute } from "react-router";
import { ROUTE_NAME_404_NOT_FOUND } from "gluestick-shared";

import MasterLayout from "components/MasterLayout";
import HomeApp from "containers/HomeApp";
import NoMatchApp from "containers/NoMatchApp";

const loadApp = (store, httpClient) => (nextState, replace, next) => {
  httpClient.get("http://localhost:7000/api/test").then((res) => {
    store.dispatch({
      type: "LOAD_DATA",
      payload: res.data,
    });
    next();
  });
};

export default function routes (store, httpClient) {
  return (
    <Route name="app" component={MasterLayout} onEnter={loadApp(store, httpClient)} path="/">
      <IndexRoute name="home" component={HomeApp} />
      <Route name={ROUTE_NAME_404_NOT_FOUND} path="*" component={NoMatchApp}/>
    </Route>
  );
}
```

### DIFF
--- a/src/lib/server/RequestHandler.js
+++ b/src/lib/server/RequestHandler.js
@@ -7,9 +7,10 @@ import Oy from "oy-vey";
 import { renderToString, renderToStaticMarkup } from "react-dom/server";
 import { match, RouterContext } from "react-router";
 import {
-    runBeforeRoutes as _runBeforeRoutes,
-    prepareRoutesWithTransitionHooks,
-    ROUTE_NAME_404_NOT_FOUND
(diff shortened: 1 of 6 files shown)


toddw:
@olya646 nice work! There was a little more work in order to make this work for server side requests as well so I have added another commit that takes care of that part.

christinebrass:
👍

Issue #427, opened 2017-01-19
Upgrade to webpack v2.2


michalchudziak:
How would you feel about upgrading webpack to the latest version? I think it's final version atm 😎

There is guide on how to migrate from version 1.x
https://webpack.js.org/guides/migrating/

toddw:
At the time of writing this ticket, I believe webpack 2.2 was the latest. I'll update the title to update to latest webpack :)

toddw:
Oh sorry. I assumed we created this ticket and you were asking for clarification. Yes, updating to the latest webpack sounds great :)

PR #456, opened 2017-01-24
Webpack Upgrade.

michalchudziak:


michalchudziak:
Solves #427.

I've migrated all the configs to match WebPack 2 API. Also `extract-text-webpack-plugin` is being updated to v2.

There are quite some changes there so i would like to invite you to discuss things you don't really like here.

package.json
```diff
@@ -60,7 +60,7 @@
     "eventsource-polyfill": "0.9.6",
     "express": "4.14.0",
     "express-pino-logger": "1.0.0",
-    "extract-text-webpack-plugin": "1.0.1",
+    "extract-text-webpack-plugin": "2.0.0-beta.5",
```
toddw, line 63: Is this the latest version? Is it possible to install 2.2.0?

src/commands/build.js
```diff
@@ -26,8 +26,15 @@ module.exports = function() {
 
   compiler.run((error, stats) => {
     const statsJson = stats.toJson();
+    const dir = "build";
+
+    if (!fs.existsSync(dir)){
+      fs.mkdirSync(dir);
+    }
```
toddw, line 33: :+1:

src/config/getWebpackClientConfig.js
```diff
@@ -10,7 +10,6 @@ const detectEnvironmentVariables = require("../lib/detectEnvironmentVariables");
 const getWebpackAdditions = require("../lib/getWebpackAdditions").default;
 const {
   additionalLoaders,
-  additionalPreLoaders,
```
toddw, line 13: it so happens that none of our apps use pre-loaders, are they no longer a thing in webpack 2?

src/config/webpack-shared-config.js
```diff
@@ -61,16 +61,21 @@ module.exports = {
...
+      }) : [
+        "style-loader",
+        "css-loader",
+        "sass-loader"
+      ],
     },
-    {
-      test: webpackIsomorphicToolsPlugin.regular_expression("json"),
-      loader: "json"
-    }
```
toddw, line 69: Is this going to break importing json files? or does that work by default in webpack 2?

src/lib/getWebpackAdditions.js
```diff
@@ -81,9 +79,8 @@ export default function (isomorphic=false) {
       additionalAliases: makeUserAdditionalAliases(additionalAliases) || {},
       additionalExternals: additionalExternals || {},
       additionalLoaders: isomorphic ? additionalLoaders : prepareUserAdditionsForWebpack(additionalLoaders),
-      additionalPreLoaders: isomorphic ? additionalPreLoaders : prepareUserAdditionsForWebpack(additionalPreLoaders),
       additionalWebpackConfig: additionalWebpackConfig || {},
-      vendor: vendor || [],
+      vendor: vendor && vendor.length > 0 ? vendor : null,
```
toddw, line 83: :+1:

templates/new/package.json
```diff
@@ -37,7 +37,7 @@
     "redux": "3.6.0",
     "sass-loader": "3.2.0",
     "style-loader": "0.13.1",
-    "webpack": "1.13.2"
+    "webpack": "2.2.0"
```
toddw, line 40: This looks right to me, is there a reason package.json at the top doesn't use this version too?

src/config/webpack-shared-config.js
toddw: Answered my own question 😅 
https://github.com/webpack/json-loader
> Note: Since webpack v2, JSON files will work by default.

src/config/getWebpackClientConfig.js
michalchudziak: Yeah now you have to add `enforce: 'pre'` to loader. Preloaders are completely gone https://webpack.js.org/guides/migrating/#module-preloaders-and-module-postloaders-was-removed

package.json
michalchudziak: That's the latest available on `npm`

toddw: If that is that is the latest that why do we have 2.2.0 down below?

toddw: Are you sure 2.2.0 is not available?

michalchudziak: image

michalchudziak: 2.0.0-rc.0 is available now tho :)

zamotany: according to this: https://github.com/webpack/webpack/releases
2.2.0 is only on github and this is the first 2.x version of webpack, previous one were a pre-resease so 2.0.0-rc.0 is the latest one on npm

templates/new/package.json
michalchudziak: Does `package.json` at the top use `webpack` dependency? I can't really see it.

toddw: I think I was mixing up extract text with webpack. Sorry for the confusion.

Issue #419, opened 2017-01-17
Declarative generator API and custom generators


zamotany:
I belive it would be great if user could write it own custom generator and then use it with GlueStick CLI (similarly to Aurelia CLI). This would require to make declarative generator API or by configs (like JSON for instance) file. User would then generate generator template, fill it/configure and use it as it would use for example `component` template.

Possible use:
1. Generate generator template
```
gluestick generate generator // or something similar like `gluestick generate --new`
```
2. Configure it
3. Use
```
gluestick generate myCustomWhatever
```

toddw:
Yeah, we talked about this in the past and I agree it would be nice to have. I'll add it to the 1.0 sprint

zamotany:
Merged

PR #458, opened 2017-01-24
Generator API + custom generators

zamotany:
See #419


src/commands/generate.js
```diff
@@ -1,154 +1,19 @@
...
-  }
-
-  if (testFileExists) {
-    return cb(`Unable to create test file for ${highlight(generatedFileName)} because it already exists`);
-  }
+const generator = require("../generator");
 
+module.exports = exports = (generatorName, entityName, options) => {
+  /**
+   * TODO: pass options from CLI to generator
```
toddw, line 6: Is this still a TODO?

src/generator/predefined/generator.js
```diff
@@ -0,0 +1,119 @@
...
+   * Define shared arguments.
+   * Those arguments that are passed down to every entry.
+   * Those arguments can be overwritten by entry-specific arguments.
+   * Type: Object
+   */
+  args: {
+    var: JSON.stringify(options)
+  },
+  /**
+   * Define modificator to alter existing files or create new one.
```
toddw: There are a handful of places that use the word `modificator`. I think we really want `modifier`?

src/generator/predefined/reducer.js
```diff
@@ -0,0 +1,74 @@
...
+
+export default combineReducers({
+  ${name}
+});
+`;
+
+module.exports = exports = options => ({
+  modify: {
+    file: "src/reducers/index",
+    modificator: content => {
```
toddw: just mentioning one more time that I think all `modificator` should be `modifier`

test/generator/createTemplate.spec.js
```diff
@@ -0,0 +1,43 @@
+const createTemplate = require("../../src/generator/createTemplate");
```
toddw: The tests files are currently named `filename.spec.js`, the convention in the project is to use `filename.test.js`

toddw:
It seems like all of the predefined generators are creating files with a blank line on top. I'm guessing this is because of how we are using
```
createTemplate`
/** … first line here… */
`;
```

It would be nice if we resolved that because I imagine a lot of people will do the same thing in their templates. Maybe we should check if the first line is blank then we remove it? What do you think?

src/generator/index.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {Object} logger Logger instance
+ */
+module.exports = exports = ({ generatorName, entityName, options }, logger) => {
+  const generator = requireGenerator(generatorName);
+  const generatorConfig = parseConfig(
+    generator.config,
+    Object.assign({}, options, { generator: generator.name, name: entityName })
+  );
+  writeTemplate(generatorConfig);
+  logger.success(`${generator.name} ${entityName} generated successfully`);
```
toddw: It would be nice to all of the files that were generated along with the path to those files. Especially with custom generators which could potentially place files in un-obvious places.

toddw:
Previously we were able to specify a folder when generating containers and components like:
```
gluestick generate component folderName/MyComponent
```

This would generate MyComponent in _src/components/folderName/MyComponent.js_ and _test/components/folderName/MyComponent.test.js_. This functionality is heavily used. It almost works in this PR but it throws an error and the output in the component needs to be updated slightly.

toddw:
`gluestick generate reducer _reducer_name_` used to not only generate the reducers but it adds an export function on _src/reducers/index.js_ so that it is automatically picked up by the store in the GlueStick app. We definitely don't want to lose that functionality.

src/generator/predefined/component.js
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+const testTemplate = createTemplate`
+/*global React*/
+/*global describe it*/
+/*global expect*/
+import ${args => args.name} from "${args => args.path}";
+import { shallow } from "enzyme";
+
+describe("${args => args.name}", () => {
```
toddw: This should be the path `components/Todos` rather than `Todos`… I realize our previous template did not include `src/` in front but I would be okay if it does now since generators can generate files all over the place.

toddw: but it shouldn't be a relative path like `../../src/components/Todos`

src/generator/predefined/component.js
```diff
@@ -0,0 +1,66 @@
...
+      path: "src/components",
+      filename: options.name,
+      template: options.functional ? functionalComponentTemplate : classComponentTemplate,
+    },
+    {
+      path: "test/components",
+      filename: `${options.name}.test.js`,
+      template: testTemplate,
+      args: {
+        path: `../../src/components/${options.name}`
```
toddw: Would be nice to take advantage of our `components` alias. The import should be able to:
```
import ComponentName from "components/ComponentName";
```

We don't need the full relative path.

src/generator/predefined/container.js
```diff
@@ -34,4 +37,42 @@ export class __$NAME__ extends Component {
...
+      path: "src/containers",
+      filename: options.name,
+      template: containerTemplate
+    },
+    {
+      path: "test/containers",
+      filename: `${options.name}.test.js`,
+      template: testTemplate,
+      args: {
+        path: `../../src/containers/${options.name}`
```
toddw: containers also has an alias so we should be able to do
`import ContainerName from "containers/ContainerName";`

toddw (changes requested):
This is *seriously* great work! I'm very impressed and very happy with this PR. There are a few very minor changes that need to be make before we can merge but excellent work!

src/commands/generate.js
zamotany: Yeah, I'm fighting with commander to allow to pass options from cli to generator, but commander is   throwing exceptions all over the place.

zamotany:
> It seems like all of the predefined generators are creating files with a blank line on top. 

Yeah it can be removed.

> Previously we were able to specify a folder when generating containers and components

Can do.

>  adds an export function on src/reducers/index.js

It's already implemented, see reducer generator

src/commands/generate.js
toddw: Should we create another ticket for tackling this or do you think it should block this PR?

toddw:
> > adds an export function on src/reducers/index.js

> It's already implemented, see reducer generator

Can you verify that it works? I saw that but it didn't seem to work in a new project. I'll try again now

zamotany:
Everything works like a charm.

src/commands/generate.js
zamotany: I will create new issue, since it more like a problem with commander. Besides it's not blocking, it's just an enhancement.

toddw:
I've tested everything out and it works really well!!! So stoked about this!

PR #459, opened 2017-01-24
fix `config` alias

toddw:
Alias were being defined in two locations. We couldn't consolodate them
into webpack-shared-config because it imported
webpack-isomorphic-tools-config, so I consoldated them there instead.

Now there is only 1 place to update alias


christinebrass:
👍

Issue #422, opened 2017-01-18
Docs setup


michalchudziak:
I have feeling that docs are quite raw. How about gathering some ideas to make it more user-friendly and informative?

ferrannp:
Right now docs point to https://www.drivenbycode.com/getting-started-with-gluestick/ which it might be / will be out of date? Maybe we can go with [GitBook](https://www.gitbook.com/) or similar.

grabbou:
Yeah. I suggest we set up Redux like approach with Gitbook powered documentation deployed to `gh-pages` by CircleCI on every pull-request.

We have that set up in place internally so it's just a matter of porting.

Can take it, curious whats @toddw take on that.

toddw:
@grabbou @ferrannp The blog post is very out of date. I would be very happy with GitBook/gh-pages. Whatever you think will have the lowest amount of friction. I feel like Redux's documentation is one of the best written that I've read.

A while back we did create something on gitbooks.com just to check it out but we never made it every far. For now it's just a copy of the README which was really just set up as a placeholder. https://www.gitbook.com/book/toddw/gluestick

I added @grabbou, I couldn't find an account to link for @ferrannp but @grabbou should have access to inviting all of you.

Is there an advantage to having it deployed to `gh-pages` vs GitBook, are those one in the same?

lukewalczak:
Can You give me and @olya646 permission to GitBook ?

grabbou:
@lukewalczak @olya646 Let's discuss that once again tomorrow as we do not need GitBook permissions to roll it out. What we can do instead is to make gitbook cli build the docs locally on the CI machine and ship them to `gh-pages`, so they are available under github.io domain (just like Redux does). I think it's better to connect documentation with the repo itself (have a top-level `docs` folder) instead of a 3rd party service that has to be checked over and over to update the docs. Also, having a local build of gitbook makes it easy for us to apply any custom styling tweaks we can think of :) 

We already have a working setup in place, will be happy to follow-up with it tomorrow.

toddw:
@grabbou This sounds great!

grabbou:
@toddw can you please turn on Github pages for the `gh-pages` branch?

toddw:
Should be set up now http://truecar.github.io/gluestick/

olga32123:
Added some docs from blog

PR #462, opened 2017-01-25
#422 Initial Gitbook impl

grabbou:
Initial Gitbook implementation.

We need to set up github pages and let Travis run `docs:publish` automatically so that we get an automatic deploy each time one changes something.


docs/Readme.md
```diff
@@ -0,0 +1,31 @@
+GlueStick
+=========
+
```
grabbou, line 3: This is going to be a new refined Readme customer oriented. I propose we keep the project Readme as dev-oriented as possible with minimal content (requirements etc.) 

Selling points can go here.

grabbou: Not yet finished, but just a placeholder for further work

grabbou:
Once the above is done and we have a runnable example, we can then iterate on the conventions and how we see the docs structured.

grabbou:
Fixes #422

toddw:
Should we change the base from `develop` to `gh-pages`?

docs/Readme.md
```diff
@@ -0,0 +1,31 @@
...
+Create and deploy enterprise-grade React and Redux application right from your command line. Gluestick is a combination
+of `create-react-app` and `next.js` that is backed and used in production by TrueCar.
```
toddw, line 5: Love this description of GlueStick!

docs/contributing/TurnOffJavaScript.md
```diff
@@ -0,0 +1,2 @@
+Now…to server-side rendering. Try turning off javascript in your web browser and refreshing the page. You’ll notice that you get the exact same result! That is because we render the same page on the server before it is sent to the user. This makes it so that your React applications can be fully search engine optimized.
```
grabbou: All the files are missing `<h1>` tag:
```md
# Turn Off JavaScript
```
etc.

docs/contributing/PrefetchingWithFetchData.md
```diff
@@ -0,0 +1,31 @@
+Earlier, we mentioned the static fetchData method. React Router allows you to call a method before rendering a component. In this case, all of the containers have the opportunity to let the server and client know if you need to fetch data before showing the component.
+To use this method, create an action creator using the Promise middleware and return the result of dispatching that action.
```
grabbou: We can try to make some more spaces here and in other docs. E.g. here, you can add an empty line before `To use this` and after it.

The convention would be to keep paragraphs separate (not line by line)

Summary.md
```diff
@@ -1,5 +1,20 @@
 # Summary
 
-* [New introduction](/docs/Readme.md)
+* [Quickstart](/docs/Readme.md)
 * [Contributing]()
-    * [Docs](/docs/contributing/Docs.md)
```
grabbou: This should be `docs/contributing.md` because it describes how to contribute. It's not `installation` file. Can we change it that way?

Summary.md
```diff
@@ -1,5 +1,20 @@
...
\ No newline at end of file
+    * [Installation](/docs/contributing/Installation.md)
+    * [Creating a new application](/docs/contributing/CreatingNewApp.md)
```
grabbou: Please put this and all the other documents you added inside `development` instead of `contributing` (that will be converted to a single file, see above). We will update that anyway, but this is just for now.

grabbou (commented):
Looks very good. I just found a couple of things that I wanted to fix before this can land.

grabbou:
@toddw not sure I fully understood, but I think we don't have to since the script will push all `develop` updates to `gh-pages` on its own

grabbou:
Demo here http://truecar.github.io/gluestick/

grabbou:
(I think it can be merged as long as there are no stylistic requests to the rest of this PR, @toddw)

Issue #452, opened 2017-01-23
Remove Mocha/Karma/node-inspector/instanbul related code


toddw:
Now that we are using Jest it seems like we should remove the dead code paths for previous testing approaches. I think we should remove anything related to Mocha/Karma/node-inspector and instanbul.

ferrannp:
Mmm but isn't this depending on #447 ?

toddw:
Yeah, some of these have been done, and some are blocked but I wanted a ticket to track all of it to make sure it doesn't slip through the cracks.

PR #465, opened 2017-01-25
Move generated app to use Jest as test solution

### LINKED ISSUE #447: Move generated app to use Jest as test solution
Following #406. We can remove all other dependencies and use also Jest in generated app (recommended by community and used it in React and Redux projecst). User can also choose to use Snapshots if they want too (maybe we could pass a flag to generate snapshot test if they want to).

This will also fix #333 since we won't need istanbul anymore.

This depends on following PR: #437 & #445

ferrannp:
Fixes #447, #333, #452.

- [x] Modify `test.js` to use Jest
- [x] Configure Jest to detect webpack aliases in tests
- [x] Handle arguments (`--coverage`, `--watch`,`--pattern`)
- [x] Handle `--debug-test` option
- [x] Delete all related to `karma`, `istanbul`, `mocha`, etc


src/commands/test.js
```diff
@@ -1,171 +1,53 @@
...
-    args.push("--debug-brk");
-  }
-  if (!options.single && !options.debugTest) {
-    args.push("--watch");
-  }
-  return args;
-}
+  // We map webpack aliases from webpack-isomorphic-tools-config file so Jest can detect them in tests too
+  Object.keys(alias).forEach((key) => {
+    moduleNameMapper[`^${key}(.*)$`] = `${alias[key]}$1`;
```
toddw, line 15: :+1:

toddw:
This is looking great so far

toddw:
We should merge in the changes from https://github.com/TrueCar/gluestick/pull/458 and update the predefined generators but I don't think that should be too much work

toddw:
I just tested out `--coverage`, `--watch` and `--pattern` and they worked great.

Also tested out `--debug-test` and it worked great too! Very glad we don't even have to pass a filename to it! 👍

ferrannp:
This is ready to be merged. @toddw I tried to make sure to remove all old unused testing code but reviewer might want to double check :)

toddw:
Looks good to me. Please resolve merge conflict and fee free to merge. Thanks!

PR #467, opened 2017-01-25
Merge: #430 Fix bundle generation back into develop

### LINKED ISSUE #276: css files imported in Index.js don't seem to be included in the production bundle
Most likely this broke with the multiple entry points stuff that was added. In a sample app I spun up I couldn't reproduce but in a couple of existing apps I am able to reproduce. I'll work more details when I get more time.

toddw:
fixes https://github.com/TrueCar/gluestick/issues/276


christinebrass:
👍

PR #471, opened 2017-01-26
Expose node_modules/.bin/ directory trough `gluestick bin` command.

michalchudziak:
Now it is possible to run `gluestick bin <dependency_name> -- <options>` from the CLI to access dependency directory.

### DIFF
--- a/src/commands/bin.js
+++ b/src/commands/bin.js
@@ -1,29 +1,17 @@
-import { exec } from "shelljs";
+import { spawn } from "cross-spawn";
 import path from "path";
 
-const invokeDependency = (name) =>
+// Creates patch to dependency's bin directory
+const getDependencyPath = (name) =>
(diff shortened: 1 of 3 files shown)


grabbou:
It is worth mentioning here that the idea of using `--` is more or less our inspiration to stay idiomatic with npm scripts.

src/commands/bin.js
```diff
@@ -0,0 +1,17 @@
...
+  path.join(__dirname, "..", "..", "node_modules", ".bin", name);
+
+// `opts` is array of options with Command object attached as last element
+module.exports = function(dependencyName, ...opts) {
+  spawn(
+    getDependencyPath(dependencyName),
+    opts.splice(0, opts.length - 1),
+    {
+      stdio: "inherit",
+    }
```
grabbou, line 15: Very nice! 👍

toddw:
This works great! 👍

Issue #469, opened 2017-01-26
Pass additional options from CLI to generator


zamotany:
It should be possible to pass addition options from CLI when running sth like:
```
gluestick generate <entity> <name> --options <generator-options>
```
those generator options should be then passed to generator.
It mainly involves figuring out why commander is throwing errors, mostly:
```
TypeError: Cannot read property 'length' of undefined
```
or
```
TypeError: this.options[i].is is not a function
```
Then adding options from `--options` param to `filteredOptions` variable in `src/commands/generate.js`.
`<generator-options>` must be `key:value` pairs, so I suggest that they should have JSON format:
```
gluestick generate <entity> <name> --options {"someOpt":true}
```
and then, `JSON.parse`'d before they are included in `filteredOptions`.

PR #472, opened 2017-01-27
Added options to pass to the generator

lukewalczak:
Solves #469.

Example usage : 
`
gluestick generate <entity> <name> -O '{ "someOpt": "someValue" }'
`


toddw:
👍

Issue #488, opened 2017-01-31
gluestick-cli package


zamotany:
Write gluestick-cli package that will wrap commands from CLI:
```
gluestick <cmd> <args> [options]
```
and spawn local gluestick with those data:
```
node ./node_modules/.bin/gluestick <cmd> <args> [options]
```
and when running `new` command it should create directory and install `npm install gluestick` in it then run `node ./node_modules/.bin/gluestick new <appName>`

PR #491, opened 2017-01-31
gluestick-cli

zamotany:
Initial implementation of #488


packages/gluestick-cli/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+commander
+  .command('new')
+  .description('generate a new application')
+  .arguments('<appName>')
+  .option('-d, --dev <path>', 'development')
+  .action((appName, options) => {
+    mkdir.sync(path.join(process.cwd(), appName));
+    spawn.sync(
+      'npm',
```
toddw, line 18: would it be better to use `yarn add` instead of `npm install`?

zamotany: I'll see what i can do about yarn but using yarn will require some more work to be done, not only here

toddw: If it isn't too much trouble I think it would be worth creating a new issue for using `yarn` over `npm install` for the speed advantage.

Issue #473, opened 2017-01-30
Refactor commands: bin


michalchudziak:
I think we should only add flowtypes here.

PR #494, opened 2017-01-31
Command bin refactor

lukewalczak:
Solves #473 

Added flowtypes to bin command


ferrannp (changes requested):
@lukewalczak you need to merge `v2` into your branch and then make sure that the related test continue to work (and add new tests if you added new functionality), in this case, `bin.test.js`.

packages/gluestick/src/types.js
```diff
@@ -0,0 +1,31 @@
...
+
+export type Context = {
+ config: {
+   projectConfig?: ProjectConfig;
+   GSConfig?: GSConfig;
+   webpackConfig?: WebpackConfig;
+   plugins: Array<Object>;
+ },
+ logger: {
+   success: Function; // Function: (...args: string) => void
```
ferrannp: `Function` is a really weak type, it means it can be any type of function. Better approach would be (didn't test it):

```
type LoggerTypes = 'success'
  | 'info'
  | 'warn'
  | 'debug'
  | 'error';
```

```
success: (type: LoggerTypes) => (args: Array) => void;
// Same for others
```

ferrannp (changes requested):
Eslint is failing in this branch. We missed it because CircleCI is not running it right now.

ferrannp (approved):

PR #496, opened 2017-01-31
Fix failing Circle builds

zamotany:


packages/gluestick/package.json
```diff
@@ -74,17 +73,6 @@
     "inquirer": "3.0.1",
     "jest": "18.1.0",
     "jsdom": "9.3.0",
```
ferrannp: Can you remove also `jsdom` ?

ferrannp (changes requested):

grabbou (approved):

ferrannp (approved):

PR #498, opened 2017-01-31
Remove support for non-`v1` branches

christinebrass:
Keep building `v1` branches and `vX.X.X` tags but remove support for other `vX` type branches.


toddw:
:+1:

Issue #495, opened 2017-01-31
gluestick-cli: add command to automatically apply changes in gluestick


zamotany:
Add command to gluestick-cli package that will listen to changes in gluestick folder and apply changes using: [wml](https://github.com/wix/wml)
Relative path to gluestick from project's `package.json` should be joined with CWD - that will be source and applied to CDW + `node_modules/gluestick` (destination)

PR #504, opened 2017-02-01
Added command to automatically apply changes in gluestick

lukewalczak:
Solves #495.


packages/gluestick-cli/index.js
```diff
@@ -47,6 +47,31 @@ commander
   });
 
 commander
+  .command('watch')
+  .description('applying changes in gluestick folder')
```
zamotany: Maybe change that to: _watch and apply changes from gluestick to project_ ?

packages/gluestick-cli/index.js
```diff
@@ -47,6 +47,31 @@ commander
...
+  .action(() => {
+    let srcPath = require(path.join(process.cwd(), 'package.json')).dependencies.gluestick;
```
zamotany: This may fail if CWD is not valid project, so this should be wrapped into try/catch block and add more meaningful message like:
- _Cannot find package.json in `<process.cwd()>`_ if require fails
- _Path is not a valid GlueStick project_ if `gluestick` is not a dependency

Also let's check if this depenedency is a path, not version string with some regex.

packages/gluestick-cli/index.js
```diff
@@ -47,6 +47,31 @@ commander
...
+    srcPath = srcPath.replace('file:', '');
+    srcPath = path.join(process.cwd(), srcPath);
```
grabbou: All these 3 lines can be e.g. extracted to function. Anyway, please try to rewrite it using `const srcPath` and avoiding overriding references.

grabbou (changes requested):

grabbou (approved):

PR #507, opened 2017-02-01
Fix all Flow issues + setting it up in CircleCI

ferrannp:


packages/gluestick/.flowconfig
```diff
@@ -1,11 +1,21 @@
...
+.*/node_modules/radium/.*
 
 [include]
 
 [libs]
 flow/
 
 [options]
+suppress_type=$FlowIssue
+suppress_type=$FlowFixMe
```
michalchudziak: Why do we need all of these ? Isn't $FlowFixMe enough :D?

PR #509, opened 2017-02-01
Restructure project test directories

### LINKED ISSUE #482: Change test structure for Gluestick project
We think that different structure for tests would be amazing - keeping `__tests__` folder in each directory will make it better for huge projects.

This is the approach recommended (and optional) by Jest. You can see this convention used in [React codebase](https://github.com/facebook/react/tree/master/src).

ferrannp:
Fixes #482 

- [x] Gluestick project: Move tests to `src` and use `__tests__` convention (Gluestick project)


packages/gluestick/templates/new/src/config/__tests__/_entry.test.js
```diff
@@ -1,9 +1,9 @@
-import React from 'react';
-import { Route } from 'react-router';
-import { shallow } from 'enzyme';
-import { getHttpClient } from 'gluestick-shared';
+import React from "react";
```
michalchudziak, line 1: Singlequote

michalchudziak (changes requested):

michalchudziak (approved):

packages/gluestick/templates/new/src/config/__tests__/_entry.test.js
ferrannp: Yup, this is because the `.eslintrc` from `templates`. We should create an issue to refactor it.

PR #510, opened 2017-02-01
add src/index.js to addon regex

toddw:
gluestick-addon-* prefix is used for components that need not be pre-compiled because they will be used from within a GlueStick project. We will compile code when webpack does it's thing inside of a GlueStick app for code in the `gluestick-addon-X/src` folder. This update includes `gluestick-addon-X/index.js`


christinebrass:
👍

Issue #484, opened 2017-01-30
Refactor autoUpgrade


michalchudziak:
- Also ask about https://github.com/TrueCar/gluestick/blob/develop/src/autoUpgrade/previous/.babelrc

PR #517, opened 2017-02-02
Refactor autoUpgrade

michalchudziak:
Solves #484


ferrannp (changes requested):
@mike866 you are missing to unskip the `updatePackage.test.js` and probably refactor it too.

Also, do you know why is the `previous` folder for ?

It also looks like it broke `flow`.

michalchudziak:
@ferrannp Yeah it's still in progress :) Also I've asked @toddw about previous folder.

toddw:
Regarding: _.../previous/.babelrc_:
> That file can actually be removed. It was once used as part of upgrade functionality we had for the .babelrc file
> But it is no longer used

It should have been removed with https://github.com/TrueCar/gluestick/commit/99fe1614e3a252b4a4a91f99726e23c2fe9e3ef0#diff-56dac13b02c65e3b185373351e13081d but it slipped past us

ferrannp (approved):

PR #519, opened 2017-02-02
[V1] Broken generate command

ferrannp:
- Was throwing an error if you do not pass options
- Written a test to be sure this won't happen again


toddw:
:+1:

PR #520, opened 2017-02-02
[V1] Update generator test template to use Jest

ferrannp:
- Generated tests with `generate` command will not work without these changes


toddw:
:+1:

PR #529, opened 2017-02-03
Refactor test command

ferrannp:


zamotany (changes requested):
Can you replace ES5 functions with arrow functions??
We want to keep consistency and in every file we refactor/write we use arrow functions.

zamotany (approved):

PR #536, opened 2017-02-06
[v1] Fixed debug server

### LINKED ISSUE #531: [v1] error: unknown option `--debug-server` (gluestick start --debug-server)
## Bug
It does not currently work as expected. It appears the option is not being passed from `start` to `start-server`.

`gluestick start --help` shows that `--debug-server` should be a valid option.

### Reproduction Steps
1. Run `gluestick start --debug-server` or `gluestick start -D`

### Result
```
error: unknown option `--debug-server`
```

### Expected Result
It should pass the option to start-server and allow for debugging server side rendering. This works as expected when you run `gluestick start-server` alone.

`gluestick start-server --debug-server` does work as expected.

lukewalczak:
Fixed #531 .

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -265,12 +265,12 @@ async function startAll(options) {
   // true.  We only want to start the client in development mode or if
   // skipBuild is not specified
   if (!(isProduction && options.skipBuild)) {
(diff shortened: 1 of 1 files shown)


toddw:
Tested and it works 🎉  :)

Issue #530, opened 2017-02-03
[v1] support for folders in generator not working


toddw:
When we were working on refactoring generators it seems as though we may have broken the ability to generate something _into_ a folder.

### Example:
`gluestick generate component used/VehicleDetails`

### Expected Result:
This should create the following files:
_src/components/used/VehicleDetails.js_
_test/components/used/VehicleDetails.js_

### Current Result:
```
[GlueStick] ERROR: Invalid name specified
```

Is this functionality supported in a different way?

Here is a comment left on PR when I first noticed this was broken: https://github.com/TrueCar/gluestick/pull/458#issuecomment-274995877

zamotany:
Regex in `src/generator/index.js` doesn't accept `/` and throws error.
To fix it: modify regex in line 13 like so:
```
if (!/^[a-zA-Z0-9\/]+$/.test(entityName)) {
```
I forgot that `/` can also be used.

PR #543, opened 2017-02-07
[v1] Fixed support for generator

lukewalczak:
Solves #530.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "gitbook-plugin-anchorjs": "1.1.1",
     "gitbook-plugin-edit-link": "2.0.2",
     "gitbook-plugin-github": "2.0.0",
-    "gitbook-plugin-prism": "2.0.2",
+    "gitbook-plugin-prism": "2.1.0",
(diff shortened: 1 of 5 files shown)


toddw:
This gets rid of the initial error message but the generated test still needs to update the path to the component.

## Reproduction Steps:
```
gluestick new gs-folder-test && \
cd gs-folder-test && \
gluestick generate component used/Vehicle && \
gluestick test
```

## Current Result
Test fails because it is looking for _components/Vehicle_ when it should be _components/used/Vehicle_

toddw (changes requested):
This gets rid of the initial error message but the generated test still needs to update the path to the component.

## Reproduction Steps:
```
gluestick new gs-folder-test && \
cd gs-folder-test && \
gluestick generate component used/Vehicle && \
gluestick test
```

## Current Result
Test fails because it is looking for _components/Vehicle_ when it should be _components/used/Vehicle_

toddw:
works great now! :+1:

PR #544, opened 2017-02-07
Webpack configs

zamotany:
- [x] client dev config
- [x] client prod config
- [x] server dev config
- [x] server prod config
- [x] server entrypoint
- [x] flow
- [x] webpack dev
- [x] server debug
- [x] pm2 
- [x] react-hot-loader
- [x] tests
- [x] auto generate client entry init files
- [x] update new project templates
- [x] fix test command
NOTES: this PR does not contain, hook or any option to configure how gluestick works, it just webpack config with half-way done commands (they require only to write test and add flow type) and some base structure.


toddw:
I really like the direction this is going! :+1:

packages/gluestick/.babelrc
```diff
@@ -1,23 +1,16 @@
...
-    },
     "test": {
+      "presets": [
+        "react",
+        "es2015",
+        "stage-0"
+      ],
+      "plugins": [
+        "transform-decorators-legacy",
+        "closure-elimination"
```
ferrannp: Can you check if we need this `closure-elimination` for Jest?

packages/gluestick/src/config/webpack/__tests__/buildEntries.test.js
```diff
@@ -0,0 +1,58 @@
...
+  },
+  '/home': {
+    component: 'path/to/home/component',
+    routes: 'path/to/home/routes',
+    reducers: 'path/to/home/reducers',
+  },
+}), { virtual: true });
+jest.mock('glob', () => ({
+  sync: jest.fn(() => ['path/to/main/entry.js']),
+}));
```
ferrannp: You can extract all these mocks in some file and reuse it in other tests like `buildServerEntries.test.js`.

packages/gluestick/src/config/webpack/deepCopy.js
```diff
@@ -0,0 +1,14 @@
+const deepCopy = (obj: Object) => {
```
ferrannp: Wouldn't be better to use existing implementation ? Like `cloneDeep` from lodash? Then we do not need to re-do it and re-test it.

ferrannp (changes requested):

packages/gluestick/src/types.js
```diff
@@ -30,35 +61,34 @@ export type LoggerTypes = {
...
-  config: any,
+  name: string;
+  config: any;
 }
 
 export type WrittenTemplate = {
-  name: string,
-  config: any,
+  name: string;
+  config: any;
```
ferrannp, line 86: Can we avoid using `any` somehow ?

packages/gluestick/src/cli/execWithConfig.js
```diff
@@ -15,15 +15,17 @@ const execHooks = (context, hooks) => {
 module.exports = (
   func,
   commandArguments,
-  { useGSConfig, useWebpackConfig, skipProjectConfig } = {},
+  { useGSConfig, useWebpackConfig, skipProjectConfig, skipEntryGeneration } = {},
```
michalchudziak, line 33: How about typing this one 😇?

packages/gluestick/src/commands/start-server/runWithPM2.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const start = ({ config, logger }, name, entryPointPath, args) => {
+  const pm2Config = {
+    script: entryPointPath,
+    name,
+    args,
+    cwd: process.cwd(),
+    exec_mode: 'cluster',
+    instances: process.env.MAX_INSTANCES || (process.env.NODE_ENV === 'production' ? 1 : 1),
```
michalchudziak: Why `(process.env.NODE_ENV === 'production' ? 1 : 1)`? I think it's not so valid 💃

packages/gluestick/src/commands/start.js
```diff
@@ -1,27 +1,21 @@
...
 const autoUpgrade = require('../autoUpgrade');
 
 /*
  * Start server and (optionally) tests in different processes.
  *
- * @param {object} options        Command-line options object directly from Commander
+ * @param {object} options Command-line options object directly from Commander
  */
-module.exports = async (config, logger, options) => {
+module.exports = exports = async ({ config, logger }, options) => {
```
michalchudziak: Let's stick to `module.exports = () => {}`

packages/gluestick/src/config/compileGlueStickConfig.js
```diff
@@ -1,6 +1,18 @@
+/* @flow */
+
+import type { Plugin, GSConfig } from '../types';
+
 const defaultConfig = require('./defaults/glueStickConfig');
 
-module.exports = (plugins) => {
-  const GSConfigOverrides = plugins.map(plugin => plugin.body.GSConfig).filter(config => !!config);
-  return Object.assign({}, defaultConfig, ...GSConfigOverrides);
+module.exports = exports = (plugins: Plugin[]): GSConfig => {
```
michalchudziak: Same as above.

packages/gluestick/src/types.js
zamotany: This should be done in separate issue, besides I didn't wrote those types, I only replaced commas with semicolons.

packages/gluestick/src/generator/templates/HomeApp.js
```diff
@@ -1,47 +1,44 @@
...
-    import { connect } from "react-redux";
-    import Helmet from "react-helmet";
+module.exports = (createTemplate: CreateTemplate) => createTemplate`
+import React, { Component } from "react";
+import { bindActionCreators } from "redux";
+import { connect } from "react-redux";
+import Helmet from "react-helmet";
 
-    import Home from "components/Home";
+import Home from "../components/Home";
```
michalchudziak, line 10: We are not supporting aliases anymore?

michalchudziak (changes requested):

packages/gluestick/src/generator/templates/HomeApp.js
zamotany: We do but `component` alias does not make any sense. Is it shared components or components from which app?

michalchudziak (approved):

ferrannp (approved):

PR #545, opened 2017-02-07
[v1] Fix .on listener in newest node.

### LINKED ISSUE #532: [v1] autoUpgrade throws error
# Bug
An error is thrown when running `gluestick start` on an app with outdated dependencies, after agreeing to the upgrade. Yarn seems to do it's install and then an error is thrown.

### Reproduction Steps
1. `gluestick new gs-test123`
2. `cd gs-test123`
3. edit package.json and change the version of any of the packages to an older version, like `webpack 2.2.0` to `webpack 2.0.0`
4. `gluestick start`
5. Agree to prompt

### Current Result
Once yarn finishes with the install the following error is thrown:
```(node:56802) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: installProcess.on is not a function```

### Expected Result
No error to be thrown

michalchudziak:
Fixes #532.

In newest `node` we cannot really attach `.on()` a listener to the synchronous method.

So in my solution, I've dropped `.on('close')` listener and let instructions from callback to be called after `installDeps()` function call.

Other solution would be to change `spawn.sync` to `spawn` call which is asynchronous, `await installDeps()` and then attach `.on('close')` listener. 

@toddw please let me know if you like the second solution more, or merge the stuff if ti's ok.

Both should behave the same.

### DIFF
--- a/src/autoUpgrade/updatePackage.js
+++ b/src/autoUpgrade/updatePackage.js
@@ -192,11 +192,10 @@ function performModulesUpdate (mismatchedModules, done) {
   fs.writeFileSync(PROJECT_PACKAGE_LOCATION, JSON.stringify(projectPackageData, null, "  "), "utf8");
 
   cleanDeps();
-  const installProcess = installDeps();
-  installProcess.on("close", () => {
-    logger.info("node_modules have been updated.");
-    done();
-  });
+  installDeps();
+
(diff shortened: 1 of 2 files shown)


toddw:
This works great! I have no problem with this solution :+1:

PR #550, opened 2017-02-07
include user extensions in file mock regex

toddw:
GlueStick provides the ability to add app specific loaders from within an app. The extensions for these loaders exists on the `other` key inside of https://github.com/TrueCar/gluestick/blob/develop/src/config/webpack-isomorphic-tools-config.js#L59-L61

This PR will ensure file mocking continues to work on tests where one of these files was loaded.


JoeCortopassi:
👍

PR #552, opened 2017-02-08
use es2015-native-modules for better tree shaking

toddw:
https://medium.com/modus-create-front-end-development/webpack-2-tree-shaking-configuration-9f1de90f3233#.qamyygi27

> Webpack 2 now knows how to work out the imports natively, without having to convert them to CommonJS modules. This is very important because CommonJS exports leaves references alive, meaning everything you declared in exports will have to be included in your final build.


christinebrass:
👍

zamotany:
This is already implemented.

PR #556, opened 2017-02-08
Refactor new command.

michalchudziak:


ferrannp (changes requested):
- It looks like the old tests templates for generated app (with `mocha`) got into this PR. We need to replace them again for the Jest ones.

- I also have problems with the path when generating a new app.

package.json
```diff
@@ -12,6 +12,7 @@
     "flow": "lerna exec npm run flow"
   },
   "dependencies": {
+    "gluestick": "file:packages/gluestick",
```
zamotany: What is that for?

packages/gluestick/src/generator/templates/package.js
```diff
@@ -0,0 +1,67 @@
...
+      "name": "AppName",
+      "version": "1.0.0",
+      "description": "",
+      "main": "index.js",
+      "scripts": {
+        "start": "gluestick start",
+        "test": "gluestick test"
+      },
+      "dependencies": {
+        "gluestick": "../gluestick/packages/gluestick",
```
zamotany: Should not be hardcoded. Value of gluestick should be dependent on value from CLI - `--dev <path>`.

zamotany (changes requested):

ferrannp (approved):

zamotany (approved):

PR #559, opened 2017-02-08
Extends path from dependency to automatically apply changes

lukewalczak:
Extends and solves #495

### DIFF
--- a/packages/gluestick-cli/cli.js
+++ b/packages/gluestick-cli/cli.js
@@ -56,14 +56,22 @@ commander
       if (!/file:.*/.test(packageContent.dependencies.gluestick)) {
         exitWithError('Gluestick dependency does not contain valid path');
       }
-      srcPath = path.join(
-        process.cwd(),
-        packageContent.dependencies.gluestick.replace('file:', ''),
-      );
+      const pathFromDependency = packageContent.dependencies.gluestick.replace('file://', '').replace('file:', '')
+      srcPath = pathFromDependency[0] !== '/'
+        ? path.join(
+            process.cwd(),
+            pathFromDependency
+          )
+        : pathFromDependency
     } catch (error) {
       exitWithError('Invalid package.json');
     }
     const wmlBin = path.join(__dirname, './node_modules/.bin/wml');
+    spawn.sync(
+      wmlBin,
+      ['rm', 'all'],
+      { stdio: 'inherit' },
+    );
     spawn.sync(
       wmlBin,
       ['add', srcPath, path.join(process.cwd(), 'node_modules/gluestick')],
--- a/packages/gluestick-cli/cli.js
+++ b/packages/gluestick-cli/cli.js
@@ -56,13 +56,13 @@ commander
       if (!/file:.*/.test(packageContent.dependencies.gluestick)) {
         exitWithError('Gluestick dependency does not contain valid path');
       }
-      const pathFromDependency = packageContent.dependencies.gluestick.replace('file://', '').replace('file:', '')
+      const pathFromDependency = packageContent.dependencies.gluestick.replace('file://', '').replace('file:', '');
       srcPath = pathFromDependency[0] !== '/'
         ? path.join(
             process.cwd(),
-            pathFromDependency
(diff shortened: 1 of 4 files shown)


zamotany (approved):

zamotany:
You need to investigate why files from `node_modules` are copied to destination even thought `node_modules` is marked as ignored. This is really important because it will fu*k up whole gluestick dependency in project.

lukewalczak:
To avoid case described above replaced watchman with chokidar.

packages/gluestick-cli/cli.js
```diff
@@ -66,27 +68,68 @@ commander
...
+      ignored: [
+        /(^|[/\\])\../,
+        /.*node_modules.*/,
+      ],
+      persistent: true,
+    });
+    watcher
+      .on('ready', () => {
+        console.log(chalk.blue('Watching for changes...'));
+        watcher.on('add', filePath => {
```
zamotany: This event handler should be separate function because handler from `change` event look almost the same, so this function should accept argument eg.`type` that will be used in console.log like `added` or `changed`.

packages/gluestick-cli/cli.js
```diff
@@ -66,27 +68,68 @@ commander
...
+          const destPath = convertFilePath(filePath);
+          fs.copy(filePath, destPath, (err) => {
+            if (err) {
+              console.error(chalk.red(err));
+            } else {
+              console.log((chalk.green(`${filePath} -> ${destPath} [added]`)));
+            }
+          });
+        });
+        watcher.on('change', filePath => {
```
zamotany: This is almost identical to `add` event handler, so this should be extracted to separate function - see above.

packages/gluestick-cli/cli.js
```diff
@@ -66,27 +68,68 @@ commander
...
+          const destPath = convertFilePath(filePath);
+          fs.copy(filePath, destPath, (err) => {
+            if (err) {
+              console.error(chalk.red(err));
+            } else {
+              console.log((chalk.green(`${filePath} -> ${destPath} [changed]`)));
+            }
+          });
+        });
+        watcher.on('unlink', filePath => {
```
zamotany: This handler should be extracted to separate function, see `addDir` and `unlineDir` handlers they look almost the same, except for `console.log`.

packages/gluestick-cli/cli.js
```diff
@@ -66,27 +68,68 @@ commander
...
+          const destPath = convertFilePath(filePath);
+          fs.remove(destPath, (err) => {
+            if (err) {
+              console.error(chalk.red(err));
+            } else {
+              console.log(chalk.green(`${destPath} [removed]`));
+            }
+          });
+        });
+        watcher.on('addDir', filePath => {
```
zamotany: Should be extracted to separate function - see above.

packages/gluestick-cli/cli.js
```diff
@@ -66,27 +68,68 @@ commander
...
+          const destPath = convertFilePath(filePath);
+          fs.ensureDir(destPath, (err) => {
+            if (err) {
+              console.error(chalk.red(err));
+            } else {
+              console.log(chalk.green(`${destPath} [added dir]`));
+            }
+          });
+        });
+        watcher.on('unlinkDir', filePath => {
```
zamotany: Should be extracted to separate function - see above.

packages/gluestick-cli/package.json
```diff
@@ -35,8 +35,10 @@
     "babel-preset-es2015": "6.22.0",
     "babel-register": "6.22.0",
     "chalk": "1.1.3",
+    "chokidar": "^1.6.1",
     "commander": "2.9.0",
     "cross-spawn": "5.0.1",
+    "fs-extra": "^2.0.0",
     "mkdirp": "0.5.1",
     "wml": "0.0.82"
```
zamotany: Don't need it, should be removed.

zamotany (changes requested):

zamotany (approved):

PR #566, opened 2017-02-09
[V1 -> V2] Generate genOptions fix + Flow

ferrannp:


packages/gluestick/src/commands/generate.js
```diff
@@ -1,6 +1,21 @@
+/* @flow */
+
+import type { Context } from '../types';
+
 const generator = require('../generator');
 
-module.exports = ({ config, logger }, generatorName, entityName, options) => {
+type GeneratorName = 'conatiner'
```
zamotany: Generator name can be any valid string, since there is possibility to create custom generator.

zamotany (approved):

PR #568, opened 2017-02-10
Remove package.json(s) for new.test.js and virtually mock them

ferrannp:
For this PR to work, we need new version of Jest that includes https://github.com/facebook/jest/pull/2486.

That Jest PR is merged but new version is still not released.


ferrannp:
Jest 19 is ready, we should be able to try it in this branch.

PR #569, opened 2017-02-13
Separate file for watch command

lukewalczak:
Extends : #495.


packages/gluestick-cli/watch.js
```diff
@@ -0,0 +1,82 @@
+const path = require('path');
+const chalk = require('chalk');
+const chokidar = require('chokidar');
+const fs = require('fs-extra');
+
+const exitWithError = message => {
```
zamotany: This function is already defined in `cli.js`, so `watchCmd` should accept it as an argument.

zamotany (changes requested):

zamotany (approved):

PR #582, opened 2017-02-15
[V2] Clean up generated project 

ferrannp:


ferrannp:
[clabot:check]

ferrannp:
[clabot:check]

packages/gluestick/src/generator/templates/AppServer.js
```diff
@@ -1,11 +1,23 @@
...
 
 module.exports = (createTemplate: CreateTemplate) => createTemplate`
-  export default {
+/* @flow */
+
+type ApplicationServer = {
+  protocol: string;
+  host: string;
+  port: number;
+  ssetPort: number;
```
zamotany: `ssetPort`?? It should be `assetsPort`

packages/gluestick/src/generator/templates/AppServer.js
```diff
@@ -1,11 +1,23 @@
...
+}
+
+const applicationServer: ApplicationServer = {
   protocol: "http",
   host: "0.0.0.0",
   port: 8888,
   ssetPort: 8880
```
zamotany: same as above

packages/gluestick/src/generator/templates/EntryWrapper.js
```diff
@@ -31,12 +33,12 @@ export default class EntryWrapper extends Component {
 }
 
 // This function is called only on client.
-EntryWrapper.start = (getRoutes, getStore, match = originalMatch, history = browserHistory) => {
+const start = (getRoutes, getStore, match = originalMatch, history = browserHistory) => {
   // Allow developers to include code that will be executed before the app is
   // set up in the browser.
   require("config/init.browser");
 
-  const httpClient = getHttpClient(config.httpClient);
```
zamotany, line 39: We might need that. I don't know what it does, but maybe it's used in TrueCar apps.

packages/gluestick/src/generator/templates/package.js
```diff
@@ -9,52 +10,42 @@ module.exports = (createTemplate: CreateTemplate) => createTemplate`
...
-    "style-loader": "0.13.1",
-    "webpack": "2.2.0"
+    "redux": "3.6.0"
   },
   "author": "",
   "license": "ISC",
   "devDependencies": {
     "babel-eslint": "7.1.1",
     "babel-jest": "18.0.0",
+    "babel-core": "6.22.1",
```
zamotany: I not sure if those deps should be in `devDependencies`, because in production gluestick still make production bundle.

zamotany (changes requested):

packages/gluestick/src/generator/templates/AppServer.js
ferrannp: @zamotany It was already like that so I wasn't paying attention to it. I'll double check because I wonder when did that changed (if before was actually `assetsPort`).

packages/gluestick/src/generator/templates/EntryWrapper.js
ferrannp: It is in the line below, I just removed `config.httpClient` which does not exist (check whole file because for some reason github is not showing this properly).

packages/gluestick/src/generator/templates/package.js
ferrannp: cc @mike866

packages/gluestick/src/generator/templates/AppServer.js
ferrannp: Ok so should be https://github.com/TrueCar/gluestick/blob/f52e1748ab8b4ce3ab0a4365371e20927c554ccd/templates/new/src/config/application.server.js#L5 `assetPort`. Since `templates` are moved in `v2`, the name has been wrong.

packages/gluestick/src/generator/templates/EntryWrapper.js
zamotany: I know that you removed `config.httpClient`. I'm also aware that in `config`, `httpClient` is not defined, but some existing apps might use it.

zamotany (approved):

PR #588, opened 2017-02-16
Better support webpack alias and loaders in mocha tests

toddw:
This is already set up in the v1 branch using Jest but this makes things better in 0.10.x in the meantime.


toddw:
[clabot:check]

src/lib/testHelperMocha.js
```diff
@@ -1,21 +1,30 @@
...
-// **ACHTUNG** THIS SHOULD ALWAYS BE RUN BEFORE ANY OTHER REQUIRE()s.
-//
-// Cannot use path.join here since it would take a require()!
-const addPath = require("app-module-path").addPath;
-addPath(CWD + (IS_WINDOWS ? "'" : "/") + "assets");
-addPath(CWD + (IS_WINDOWS ? "'" : "/") + "src");
+  const mock = RegExp(`^[./a-zA-Z0-9@$_-]+\\.(${images.join("|")}|${other.join("|")}|${styles.join("|")})$`);
+  if (mock.test(input)) {
+    return requireHacker.resolve(path.join(__dirname, "fileMock.js"));
+  }
```
christinebrass, line 24: I'm a little fuzzy on why we need to mock the files instead of ignoring them. Can we not ignore these by just falling through to the return?

toddw: falling through to the return will actually fall back to default `require` behavior

christinebrass:
👍

Issue #479, opened 2017-01-30
Refactor commands: start-client


michalchudziak:

ferrannp:
Wait till #544 is merged.

PR #590, opened 2017-02-17
Refactor start-client.

michalchudziak:
Solves #479.


packages/gluestick/src/types.js
```diff
@@ -89,3 +89,7 @@ export type WrittenTemplate = {
 export type GeneratorOptions = {
   [key: string]: ?string;
 }
+
+export type Compiler = {
+  run: Function,
```
ferrannp: I think you can make this more specific: `(error: string) => void;`.

packages/gluestick/src/commands/start-client.js
```diff
@@ -38,22 +62,23 @@ module.exports = ({ config, logger }) => {
...
     }));
 
     // TODO: add hook
-    app.listen(GSConfig.ports.client, GSConfig.host, error => {
+    app.listen(GSConfig.ports.client, GSConfig.host, (error: string) => {
       if (error) {
         logger.error(error);
         return; // eslint-disable-line
       }
+      /* $FlowFixMe */
```
ferrannp: What is Flow complaining about here ?

packages/gluestick/src/commands/start-client.js
```diff
@@ -1,34 +1,58 @@
...
+  // network path for static files: fetch all statics from webpack development server
+  publicPath: string,
+  headers: Object,
+  stats: Object,
+};
+
+module.exports = ({ config: { GSConfig, webpackConfig }, logger }: Context): void => {
+  if (!webpackConfig) {
+    logger.error('Webpack config not specified');
+    return;
```
ferrannp: In this cases, shall we call `process.exit(1);` to finish command with error? We do it in several places.

ferrannp (changes requested):

packages/gluestick/src/commands/start-client.js
zamotany: `new Error()` should be thrown instead of `return;` this will be cached in `cli/execWithConfig.js` which will exit process with code 1.

zamotany: Besides what's the point in checking if `webpackConfig` and `GSConfig` are not `null`? User can't modify context.

packages/gluestick/src/commands/start-client.js
```diff
@@ -1,34 +1,58 @@
...
+  if (!GSConfig) {
+    logger.error('Gluestick config not specified');
+    return;
+  }
 
-    // network path for static files: fetch all statics from webpack development server
-    publicPath: configuration.output.publicPath,
+  const configuration: WebpackConfigEntry = webpackConfig.client;
+  const publicPath: string =
+    (typeof configuration === 'object' && !Array.isArray(configuration)) ?
```
zamotany: What is `!Array.isArray(configuration)` for? `configuration` will be an object.

packages/gluestick/src/commands/start-client.js
ferrannp: Yea i was actually reading https://www.nczonline.net/blog/2014/02/04/maintainable-node-js-javascript-avoid-process-exit/ and probably we should thrown errors instead of following the `process.exit(1)` approach that is used in some parts of the code (+ this `process.exit` gives problems in test if we do not mock it).

packages/gluestick/src/commands/start-client.js
michalchudziak: image

I think if we merge @zamotany 's type updates it will be fixed.

ferrannp (approved):

PR #596, opened 2017-02-20
Flow tweaks for generator

zamotany:


michalchudziak (commented):
Fix conflict then LGTM

PR #600, opened 2017-02-20
Generate command with new structure

ferrannp:
- You pass now an `entryPoint` using -E if you want to use predefined generators (component, container and reducer)
- Types
- Tests


packages/gluestick/src/generator/generator.js
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @param command {Object} { generatorName, entityName, entryPoint, options } Command object
+ * @param logger {Object} logger Logger instance
+ */
+module.exports = (command: Command, logger: Logger): void => {
+  const {
+    generatorName,
+    entityName,
+    options = {},
+  } = command;
```
michalchudziak, line 34: Destruct in args definition

packages/gluestick/src/generator/generator.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const predefinedGenerators = [
+    'component',
+    'reducer',
+    'container',
+  ];
+
+  if (predefinedGenerators.find((element) => element === generatorName)) {
+    if (!isValidEntryPoint(entryPoint, logger)) {
+      return;
```
michalchudziak, line 47: Can't we throw an error here?

packages/gluestick/src/generator/generator.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  if (!/^[a-zA-Z0-9/_-]+$/.test(entityName)) {
+    logger.error('Invalid name specified');
+    return;
```
michalchudziak, line 53: Shouldn't we also kill process here?

michalchudziak (changes requested):

packages/gluestick/src/generator/generator.js
ferrannp: I think we should not kill the process (see https://www.nczonline.net/blog/2014/02/04/maintainable-node-js-javascript-avoid-process-exit/) but instead throw errors. 

In this case though, it is not like a code error but a error message (using their `logger.error`) to the user when typing the command. I think in this case it's better this way although I am not sure 100% sure.

packages/gluestick/src/generator/generator.js
ferrannp: Same comment as the previous `return`. `isValidEntryPoint` already logs errors + info.

packages/gluestick/src/generator/generator.js
ferrannp: This is probably a matter of test, for me it is easier to read the function this way.

michalchudziak (approved):

Issue #574, opened 2017-02-13
gluestick reset-hard command


zamotany:
Add `reset-hard` command to `gluestick-cli` that will run series (synchronously) of commands:
```
rm -rf node_modules/gluestick
rm -rf build
npm cache clean
gluestick reinstall-dev
```

PR #605, opened 2017-02-21
Added reset-hard command

lukewalczak:
Solves #574


packages/gluestick-cli/cli.js
```diff
@@ -41,6 +42,13 @@ commander
   });
 
 commander
+  .command('reset-hard')
+  .description('remove gluestick and build, cache clean and reinstall-dev')
+  .action(() => {
+    resetHardCmd();
```
dratwas: why not just `resetHard` ?

zamotany (approved):

PR #612, opened 2017-02-21
Auto-upgrade refactor

zamotany:


michalchudziak (approved):

packages/gluestick/src/autoUpgrade/__tests__/checkForMismatch.test.js
```diff
@@ -0,0 +1,55 @@
+jest.mock('../getSingleEntryFromGenerator.js', () => jest.fn());
```
ferrannp, line 2: We need to start also adding `/* flow */` into tests, they are ready. I already did it in several of them. The same goes for the other tests on this review.

packages/gluestick/src/autoUpgrade/__tests__/index.test.js
```diff
@@ -0,0 +1,68 @@
+jest.mock('fs', () => ({
```
ferrannp: I think we should start changing the `index.js` pattern. We probably should only use `index.js` to export the main entry point for each folder. It is specially useful for:

- Looking for files
- Jest normal and watch mode pattern, multiple `index.test.js` make this very hard right now

ferrannp (dismissed):

ferrannp (approved):

PR #616, opened 2017-02-22
check if route have set cache to true then cache it

dratwas:
Only routes which have set cache prop to true should be cached.


ferrannp (approved):
For future:

- We need to add Flow to the test too
- The test needs to have a test case when `cache` is false

PR #644, opened 2017-02-23
Cache tests + flow 

dratwas:


packages/gluestick/src/renderer/__tests__/render.test.js
```diff
@@ -1,3 +1,17 @@
+/* @flow */
+import type {
```
ferrannp: Can you add blank line between flow comment and import ?

packages/gluestick/src/renderer/__tests__/render.test.js
```diff
@@ -141,4 +215,99 @@ describe('renderer/render', () => {
       });
     });
   });
+
+  describe('without a custom render method and cache set to false', () => {
```
ferrannp: Mmmm all these tests seem exactly the same copy as when cache is set to `true` but changing this case:

```
expect(cacheManager.setCacheIfProd.mock.calls.length).toBe(0);
```

You should write a function for the dynamic part so we do not repeat so much code in here.

`it`s can be dynamic.

ferrannp (changes requested):

ferrannp (approved):

PR #649, opened 2017-02-23
ssr hooks

dratwas:
Solves #614 
For working with SSR hooks we need to create a file called gluestick.hooks.js in src project directory which should contain hooks. I've added template etc so gluestick should create this file in autoUpdate or gluestick new.


michalchudziak (commented):
LGTM

zamotany (approved):

ferrannp (approved):

Issue #656, opened 2017-02-24
New script to integrate lint, flow and test-coverage


lukewalczak:
`npm run ci` instead of `npm run lint` && `npm run flow` && `npm run test-coverage`

PR #657, opened 2017-02-24
Integrate flow, lint and test

lukewalczak:
Solves #656

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "test": "lerna exec npm run test",
     "test-coverage": "lerna exec npm run test-coverage",
     "lint": "lerna run lint",
-    "flow": "lerna exec npm run flow"
+    "flow": "lerna exec npm run flow",
+    "ci": "npm run lint && npm run flow && npm run test-coverage"
   },
   "dependencies": {
     "lerna": "2.0.0-beta.34",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "test": "lerna exec npm run test",
     "test-coverage": "lerna exec npm run test-coverage",
     "lint": "lerna run lint",
-    "flow": "lerna exec npm run flow"
+    "flow": "lerna exec npm run flow",
+    "ci": "npm run lint && npm run flow && npm run test-coverage"
```
ferrannp, line 13: You need:

```
"babel-preset-es2015": "6.22.0",
"babel-preset-stage-0": "6.22.0",
```

as devDep.

packages/gluestick-env-expose-ast/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "test": "jest",
     "test-coverage": "jest --coverage",
-    "lint": "eslint ./",
+    "lint": "eslint ./*.js ./__tests__/**/*.js",
```
ferrannp: Use `eslint .` and then create a `.eslintignore` that ignores all `/coverage` folders.

ferrannp (changes requested):

packages/gluestick-env-expose-ast/package.json
zamotany: `.eslintignore` should be in root folder, next to `.eslint` because `coverage` folder will be in every package

zamotany (approved):

PR #668, opened 2017-02-27
Add deployment section to Circle CI

zamotany:
`NPM_TOKEN` needs to be set in project settings on Circle CI.
Also branch name on which deployment can be triggered probably will need to be changed.


ferrannp:
@zamotany what's the status of this? WIP?

zamotany:
Yes, it's WIP still since, there is not clear answer to have deployment on circle CI or locally and NPM_TOKEN is not set in settings on Circe CI.

Issue #460, opened 2017-01-25
Make a better Home example when generating a new project


toddw:
Currently, when you generate a new gluestick project, start it up and open in the browser you are greeted with:
```
Home
```

I think we should consider making the Home component a little more impressive. One one hand, I'm certain they are going to wipe it away. I don't want to create a lot of work for someone to _remove_ stuff we did just to make the project excite new users. On the other hand, it is very anticlimactic when you open your browser to http://localhost:8888/

This is what I propose:
We create an exciting/interesting component outside of the Home component like:
_src/components/WelcomeToGlueStick.js_

Then in the demo home project, instead of `<div>Home</div>` we have `<div><WelcomeToGlueStick /></div>`

This way, it is just as easy to remove the welcome message as it is to remove `Home`. Well, almost, they may want to delete the component too but I think it would be worth it.

As far as what WelcomeToGlueStick should have… I think it should be a decently designed page that explains a number of the features you get with GlueStick. Almost like a super abbreviated version of a GettingStarted guide.

What does everyone think?

toddw:
This probably goes without saying but we don't want to have to include any third party modules for building _WelcomeToGluestick_. We don't want to add things to the dependency tree just for the welcome message.

michalchudziak:
Yeah - how would you feel about `Welcome to GlueStick!` header with some basic usage info and simple styles?

Do you have anything on your mind which you want to be present on welcome page?

michalchudziak:
I think we could list some basic commands with explanation like: 
- `gluestick start` + description
- `gluestick test` + description
- etc...

toddw:
I love what Rails has but I think we can do better with some helpful basic commands.
![image](https://cloud.githubusercontent.com/assets/140000/22317451/9272a69c-e328-11e6-97c3-a62eb329880b.png)

I might be able to get our designers to do something for us

toddw:
I just talked to one of our designers Robby. He is going to put something together 😁 🎉

michalchudziak:
Any luck with getting designs 😇?

toddw:
I will ask Robby today. I haven't seen anything yet

toddw:
Maybe we can implement this basic thing I threw together while we wait for a fancy design :)
![gluestick-welcome-screen](https://cloud.githubusercontent.com/assets/140000/22749327/c74df632-ede1-11e6-8cc1-214d2ab36774.png)

zamotany:
[...]

PR #680, opened 2017-02-28
Make a better home example

dratwas:
Solves #460 
Now it looks like this:
screen shot 2017-02-28 at 14 20 03

In this PR I added alias called `compiled` for assets which are loaded from node_modules. I did it because universal-webpack loads assets from node_modules as .js files until those files aren't added to `exclude_from_externals` array.
Check:
https://github.com/halt-hammerzeit/universal-webpack/blob/master/README.md#advanced-configuration

We don't know which libraries/assets user want to load from node_modules so instead of adding it to `exclude_from_externals` we can use a alias which is set to node_modules path.

Now for loading i.e. normalize.css:
```javascript
import 'compiled/normalize.css';
```


ferrannp (commented):
What do you think of making it mobile friendly @dratwas ? I think we would just need the last row to be of 1 column instead on 3 on smaller screens.

ferrannp (changes requested):
Let's request the change so it is clearer to track it.

dratwas:
@ferrannp I made it mobile friendly in last commit.

ferrannp (changes requested):
- Fonts are not looking good in header and other places
- Header should has no margin

Issue #379, opened 2016-12-19
Write plugin for Radium


toddw:
I know it is possible to use Aphrodite with GlueStick but it would be nice if it were easier to set up. Currently you have to override `renderMethod` in _src/config/application.server.js_ then you have to run through the `StyleSheetServer.renderStaticMethod` as documented in the aphrodite docs which gives you the html and the css then you expose the html as `bodyContent` in the output of the render method and you have to set up the stylesheet head tag with the css exported from `StyleSheetServer.renderStaticMethod` and then you need to add code to `src/config/init.browser.js` to rehydrate aphrodite before the app is mounted.

It's all possible and we have done it but it would be nice if it were easier.

toddw:
It would be nice to have a render-style interface and then you would have a Radium style interface an aphrodite style interface or people could write their own.

michalchudziak:
Do you have any concepts on how should it look like in your mind?

Also, I've seen `Radium ` integration is being removed in this PR https://github.com/TrueCar/gluestick/pull/386 - what's the status here, should we proceed with work here?

toddw:
386 is a start, you can definitely pick up from where that one leaves off if it helps but it really only does some (possibly all) of the work for removing Radium but none of the work to make Radium or Aphrodite easy to turn on.

The way I think this should work is that a developer can specify a style engine in src/config/application.js

{..., styleEngine: gluestickRadiumEngine, ...}

And things should work with Radium. I don't know if we want to call it an engine or an interface.

What do you think about this approach? Definitely open to suggestions.

michalchudziak:
Do you think that it is that important to have style engine integrated into GlueStick? I think it's not that bad if we let the user handle it by himself, but that's just an idea. If you really want to have it integrated then I like your proposal, but I'm afraid that Radium, Aphrodite or StyledComponents have a bit different API, and it might be tricky to have it universal.

toddw:
My goal is to make the simple stuff easy and the complex stuff possible… I do believe there is a way we can pull this off even if it will be tricky.

lukewalczak:
What do you think about an idea to give user option argument to `new` command to choose `radium` or `aphrodite`?

toddw:
Allowing the user to specify the style engine in the `new` command sounds good to me, as long as it doesn't make it difficult for people to write their own style engine.

I'm going to spend some time today planning some possibilities for this functionality today. After thinking about it some more I have an idea of what might make this pretty easy to implement and give us everything we need.

[...]

PR #682, opened 2017-02-28
Radium plugin 

lukewalczak:
Solves #379


ferrannp:
@lukewalczak @zamotany  Could you make a summary of what this PR includes? It seems to include much more stuff than just writing the Radium plugin. Because I am hoping that writing a plugin does not require 18 commits and 26 changes 😆

zamotany (approved):

zamotany:
To support plugins, that are used not in config compilation phase, but in app code we needed to make some changes to gluestick, to allows that.

michalchudziak (approved):

Issue #686, opened 2017-02-28
[v2] gsBeforeRoute no longer works


toddw:
`gsBeforeRoute` doesn't actually do anything anymore.

If a static function exists on a top level container that is referenced by a route, it used to run that function before rendering. This would run on the server side or in the browser depending on how you get there. Not only that but if it returns a promise it would wait for the promise to get resolved before continuing.

I think the following things were just never hooked up:
https://github.com/TrueCar/gluestick-shared/blob/master/src/lib/route-helper.js
https://github.com/TrueCar/gluestick-shared/blob/master/src/components/TransitionHooks.js

toddw:
I'm also happy to rename this to something better… any suggestions?

dratwas:
I didn't change the name because I won't touch gluestick-shared.  We need to remember that the hooks will not be called on init. Those hooks will be called only when we change a route.

toddw:
Testing this out and it doesn't seem to be working

toddw:
Example project https://github.com/toddw/gluestick-v1-testing

zamotany:
This is not working or `rc-3` but working on `next` branch. Once #690 is settled out we will release new version.

toddw:
I tried it on the next branch and it didn't seem to be working. Can you try the example project I posted

zamotany:
I used your example project:
1. modified `package.json` and set `gluestick` to (in my case) `../gluestick/packages/gluestick`
2. `gluestick reinstall-dev`
3. `gluestick start`
then removed code in `componentDidMount` from `TodoApp` (gluestick rebuilded that).
Next, added some elements to list and refreshed page, even I opened it in incognito mode and I could see the list.

toddw:
I wasn't seeing the update because I was updating gluestick-cli and not the gluestick package. Still getting used to the new setup. Thanks for verifying. publishing 1.0.0-rc.4 now

PR #689, opened 2017-03-01
call gsBeforeRoute function from components before render

dratwas:
Solves #686


zamotany (commented):
Why are code dublicated?? Looks like `getRouteComponents` is the same as in `gluestick-shared` package.

zamotany (changes requested):
> This would run on the server side or in the browser depending on how you get there

it should run both on client (in browser) and on server - now it only runs on server
Besides there are code that is dublicated, the one from `gluestick-shared` should be used.

dratwas:
Now I use this one from gluestick-shared. Hooks work on a client but only when moves from one route to another not on init.

zamotany (approved):

PR #691, opened 2017-03-01
Not check gluestick version on autoUpgrade

lukewalczak:
Solves #621

### DIFF
--- a/packages/gluestick/src/autoUpgrade/checkForMismatch.js
+++ b/packages/gluestick/src/autoUpgrade/checkForMismatch.js
@@ -62,7 +62,7 @@ const checkForMismatch = (
     };
   };
   Object.keys(templatePackage.dependencies).forEach((dep: string): void => {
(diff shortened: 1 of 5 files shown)


ferrannp (dismissed):
Did you take into consideration that this should **only** be done when we pass `--dev` flag ?

packages/gluestick/src/autoUpgrade/checkForMismatch.js
```diff
@@ -59,6 +62,9 @@ const checkForMismatch = (requiredPackage: ProjectPackage): Promise<UpdateDepsPr
     };
   };
   Object.keys(templatePackage.dependencies).forEach((dep: string): void => {
+    if (dev && dep === 'gluestick' && !/\d\.\d\.\d.*/.test(projectPackage.dependencies[dep])) {
```
zamotany: Replace this regex with: `\d+\.\d+\.\d+.*`

zamotany (changes requested):

michalchudziak (approved):

zamotany (approved):

Issue #701, opened 2017-03-02
Bump packages versions.


michalchudziak:
We need to update `package.json` files in our packages to match ones from `npm`.

michalchudziak:
Eventually, lets set it to `0.0.0` since `lerna` takes care of version update.

PR #704, opened 2017-03-02
Bump gluestick packages version to 0.0.0

lukewalczak:
Solves #701


zamotany (changes requested):
every package should have version set to 0.0.0, including gluestick-plugin-radium, gluestick-config-legacy, gluestick-env-expose-ast and gluestick-shared

zamotany (changes requested):
gluestick-shared should also have version set to `0.0.0`

zamotany (approved):

PR #710, opened 2017-03-03
Meta object in plugins

zamotany:


packages/gluestick/src/plugins/prepareConfigPlugins.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  logger.info('Compiling plugins:');
+  try {
+    const compiledPlugins: ConfigPlugin[] = pluginsConfig.map((value: Plugin): ConfigPlugin => {
+      const compilationResults: Object = compilePlugin(value, { logger });
+      if (compilationResults.error) {
+        throw compilationResults.error;
+      }
+      logger.success(`  ${value.name} compiled successfully`);
```
dratwas, line 53: Any reason to add those white spaces before value.name?

zamotany: It will print:
```
Compiling plugins:
  <plugin-name>
  ...
```

packages/gluestick/src/config/__tests__/compileGluestickConfig.test.js
```diff
@@ -36,17 +36,17 @@ describe('config/compileGluestickConfig', () => {
       // $FlowIgnore
       {
         name: 'testPlugin',
-        body: {
-          overwriteGluestickConfig: (config) => {
+        overwrites: {
+          gluestickConfig: (config) => {
             return Object.assign(config, { protocol: 'https' });
```
michalchudziak: Maybe let's use spread in this file :>?

dratwas (approved):

PR #712, opened 2017-03-03
Separate package for gluestick generators.

michalchudziak:
WIP


michalchudziak:
Freezing this one due to symlink idea.

michalchudziak:
We need to replace `"gluestick-generators": "file:<PATH>"` with `rc` version on next release.

packages/gluestick-generators/src/index.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
+
+module.exports = {
+  default: require('./generator'),
+  parseConfig: require('./parseConfig'),
+};
```
zamotany: It's also possible to export some function as a default, and other when using dot:
```
module.exports = () => '1';
module.exports.fn = () => '2';
```
so requiring will be:
```
const fnDefault = require('some-file');
const fnOther = require('some-file').fn;
```
but I'm fine with object also.

packages/gluestick-cli/package.json
```diff
@@ -30,14 +30,17 @@
...
+    "babel-preset-stage-0": "6.22.0",
     "babel-register": "6.22.0",
     "chalk": "1.1.3",
     "chokidar": "^1.6.1",
     "commander": "2.9.0",
     "cross-spawn": "5.0.1",
     "fs-extra": "^2.0.0",
-    "mkdirp": "0.5.1",
-    "glob": "7.1.1"
+    "gluestick-generators": "file:../gluestick-generators",
```
zamotany: Not sure if this is ok

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "name": "gluestick-generators",
+  "version": "0.0.0",
+  "description": "Package with generator API used by GlueStick",
+  "main": "src/index.js",
+  "homepage": "https://github.com/TrueCar/gluestick",
+  "bugs": "https://github.com/TrueCar/gluestick/issues",
+  "license": "MIT",
+  "scripts": {
+    "test": "jest",
+    "test-watch": "jest --watch",
```
zamotany: not needed

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "test-coverage": "jest --coverage",
+    "debug": "node --inspect --debug-brk ./node_modules/.bin/jest -i --env ./test/jestEnvironmentNodeDebug.js",
```
zamotany: also not needed

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "postpublish": "./docker/create-base-image.js",
```
zamotany: same as above

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "update-version": "npm version",
```
zamotany: same as above

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "postupdate-version": "./docker/generate-dockerfile.js"
```
zamotany: same as above

packages/gluestick-generators/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
+    "babel-polyfill": "6.22.0",
+    "babel-preset-es2015": "6.22.0",
+    "babel-preset-react": "6.22.0",
+    "babel-preset-stage-0": "6.22.0",
+    "jest": "19.0.0",
+    "mkdirp": "^0.5.1"
+  },
+  "devDependencies": {
+    "temp": "0.8.3"
```
zamotany: I believe this is not needed

packages/gluestick-generators/src/requireGenerator.js
```diff
@@ -8,14 +8,17 @@ const { convertToCamelCase, convertToKebabCase } = require('./utils');
 // $FlowFixMe
 module.createTemplate = createTemplate;
 
-const PREDEFINED: string = 'predefined';
+const PATH_TO_GLUESTICK_TEMPLATES = '../../gluestick/generator/predefined';
```
zamotany: maybe use `require.resolve` and split it on `gluestick` to get always valid path to gluestick?

packages/gluestick/package.json
```diff
@@ -53,6 +53,7 @@
     "express": "4.14.0",
     "extract-text-webpack-plugin": "2.0.0-beta.5",
     "fs-extra": "2.0.0",
+    "gluestick-generators": "file:../gluestick-generators",
```
zamotany: Also not sure about this

zamotany (changes requested):

packages/gluestick-generators/src/index.js
ferrannp: I prefer this over the `.default` :)

packages/gluestick-generators/package.json
ferrannp: Yea this is not needed. It is only used in `destroy.test.js` from `gluestick` package and it will likely go away from there soon too.

packages/gluestick-generators/src/requireGenerator.js
michalchudziak: I'm not sure about this solution, in fact it results in `Cannot find module 'gluestick'`.

zamotany: In that case extract path to gluestick templates to `getPossiblePaths` and `requireGeneratorConfig` argument and pass it from `generator.js`.

zamotany (changes requested):

Issue #717, opened 2017-03-04
Update promise-middleware


toddw:
Currently the promise middleware puts the payload on the `value` key but we should update it so that it is on `payload`

PR #727, opened 2017-03-06
Promise Middleware refactor & change name of key from value to payload

dratwas:
Solves #717


zamotany (approved):
Working fine, fix lint and it will be merged

PR #729, opened 2017-03-06
backfill https://github.com/TrueCar/gluestick/pull/704

toddw:


ferrannp:
This is not valid anymore, lerna will handle versions for us + we'll be using `staging`.

PR #738, opened 2017-03-07
Deploy script

zamotany:
This **should** work, but without actually pushing packages to npm, and docker image to Docker Hub. So the real test will be when publishing new RC.


ferrannp (changes requested):
LGTM. Can you add a `publish` npm command to run the script?

ferrannp (approved):

Issue #698, opened 2017-03-02
Improve CLI commands to take packages into account


zamotany:
Right now when installing gluestick in project gluestick-shared is downloaded from npm. This is undesired behavior, in development we want both gluestick and gluestick-shared to be installed from local dev version and both to be watched.

Ideally there should be some option to tell which packages to watch and install by eg. some regex in command arg.

edit

- [x] new command PR #708 
- [x] reinstall-dev command PR #709 
- [x] watch command PR #711 
- [x] reset-hard [WIP]
- [x] moving gs-shared PR #740 
## New
```
gluestick new <projectName> --dev ./gluestick
```
`dev` should be prefixed with `..` and `packages` should be added at the end, then by default all packages should be installed, and proper paths set for them in `package.json`
If `--dev` flag is not specfied, it should only install gluestick with version taken from gluestick-cli's `package.json`.
## Reinstall-dev
```
gluestick reinstall-dev
```
In project `package.json` should find every dependency starting with `gluestick`, which have path set instead of version string.
## Watch
```
gluestick watch
```
In project `package.json` should find every dependency starting with `gluestick`, which have path set instead of version string and start watching those packages.
## Reset-hard
```
gluestick reset-hard
```
In project `package.json` should find every dependency starting with `gluestick`, which have path set instead of version string and remove those packages from `node_modules`, remove `build` folder, `npm cache clean` and call `reinstall-dev`.

lukewalczak:
Issue was replaced with #720

PR #740, opened 2017-03-07
Moving gs-shared

lukewalczak:
Moving gs-shared to devDependencies in gluestick and add gs-shared as a dependecy in project.

Solves #698


ferrannp (changes requested):
That will not work in CircleCI because version 0.0.0 of `gluestick-shared` does not exist. We need to wait for next release candidate where `gluestick-shared` will be included and published.

Issue #732, opened 2017-03-06
[1.0.0-rc] did we lose init.server.js capability?


toddw:
Did we lose the ability to include _init.server.js_. I do see init.browser.js but it looks like we may have lost _init.server.js_

https://github.com/TrueCar/gluestick/blob/develop/src/entrypoints/server.js#L9-L16

# Notes
**This is needed in the 1.0.0-rc branch**

zamotany:
We decided that it will be in `gluestick.hooks.js` as the rest, because keeping it in separate file is pointless and involves more changes than adding it as hook. Besides it's more consistent to have every hooks in one place.

toddw:
How do you feel about adding `postStart`? or is there already something for that? There are times when people need to run a script when they run `gluestick start` before the server is initiated. That is where our init.server.js code mostly came into play. We don't really want the code to run before every render but just once when the app starts off

zamotany:
It won't run with every render. It will run right before renderer code starts. The question is should this be run once when gluestick start is called or on every **rebuild and rerun** in dev mode.

toddw:
It should run once per build

PR #756, opened 2017-03-07
add init server hook

dratwas:
This PR solves #732 . Now we use hooks for that like @zamotany wrote in #732.


ferrannp:
Can you clarify last comments from #732 ?

zamotany:
Also, I'm wondering if this should be run in renderer code or in gluestick. In dev mode should this be run only once or on every rebuild and rerun?

michalchudziak (approved):

zamotany (approved):

PR #757, opened 2017-03-07
add init server hook

dratwas:
This PR solves #732.


zamotany:
Just wondering if this should be run in renderer code or in gluestick. In dev mode should this be run only once or on every rebuild and rerun?

michalchudziak (approved):

Issue #759, opened 2017-03-07
[rc] Issues when running npm publish


ferrannp:
I have permissions to publish to `npm`. It looks like rc6 was correctly published https://www.npmjs.com/search?q=gluestick but I got different errors.

1. About docker: `sh: ./docker/create-base-image.js: No such file or directory` so docker image was not successfully created
2. Some strange error: `npm ERR! "You cannot publish over the previously published version 1.0.0-rc.6." : gluestick`

I attach [publish-error.txt](https://github.com/TrueCar/gluestick/files/825848/publish-error.txt) the full stacktrace.

Moreover, the auto commit was done https://github.com/TrueCar/gluestick/commit/70b4c206ce65bd77ca380ca0e0603f32e4c5acc1 (updating gluestick versions). But, the dependency inside `gluestick` (it has `gluestick-shared` as dependency), was **not** updated https://github.com/TrueCar/gluestick/blob/staging/packages/gluestick/package.json#L59.

zamotany:
About 1: it looks like there is `postpublish` or sth like that in gluestick that sill wants to exec `./docker/create-base-image.js` which I removed.

PR #763, opened 2017-03-08
Improve deploy script

zamotany:
Connected with #759

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -7,9 +7,6 @@ if (process.argv.length <= 2 || !/\d+\.\d+\.\d+.*/.test(process.argv[2])) {
 
 const version = process.argv[2];
 const spawn = require('cross-spawn');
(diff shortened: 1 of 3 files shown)


scripts/deploy.js
```diff
@@ -5,17 +5,52 @@ if (process.argv.length <= 2 || !/\d+\.\d+\.\d+.*/.test(process.argv[2])) {
...
+const fs = require('fs');
+
+const spawnWithErrorHandling = (...args) => {
+  const results = spawn.sync(...args);
+  if (results.error) {
+    throw results.error;
+  }
+};
+
+// Update gluestick-* dependecies in packages
```
ferrannp: Let's take a close look into this because `lerna` should handle this for us.

ferrannp (dismissed):

ferrannp (approved):

Issue #742, opened 2017-03-07
Move gluestick-shared test to use Jest


ferrannp:
This is also necessary for #722. So all packages share same test suite.

PR #765, opened 2017-03-08
Using Jest in gluestick-shared tests

lukewalczak:
Solves #742


packages/gluestick-shared/src/lib/__tests__/getHttpClient.test.js
```diff
@@ -42,9 +40,7 @@ describe('lib/getHttpClient', () => {
       return i;
     }
 
-    const create = sinon.stub().returns(getInstance());
-    create.onCall(1).returns(getInstance());
-    create.onCall(2).returns(getInstance());
+    const create = jest.fn(() => (getInstance()));
```
zamotany: Those additional parentheses around `getInstance()` are not needed.

zamotany (changes requested):
Can you also update babel dependencies to more recent version?

packages/gluestick-shared/package.json
```diff
@@ -29,16 +29,19 @@
...
-    "babel-preset-stage-0": "6.5.0",
-    "chai": "^3.5.0",
-    "mocha": "^2.4.5",
+    "babel-preset-es2015": "6.22.0",
+    "babel-preset-react": "6.23.0",
+    "babel-preset-stage-0": "6.22.0",
+    "jest": "19.0.0",
     "radium": ">=0.17.1",
     "react": ">=15.1.0",
     "react-dom": ">=15.1.0",
```
ferrannp, line 47: You need to remove `sinon` too.

ferrannp (changes requested):

ferrannp (approved):

Issue #538, opened 2017-02-06
Be able to pass any CLI option to gluestick test (Jest)


ferrannp:
Jest has all these options http://facebook.github.io/jest/docs/cli.html.

Right now we are limiting this to our debug `-D`, `--watch`, `--coverage` and our own pattern `-R` to simulate `jest -- TestName`. 

It would be nice to be able to use `gluestick test` exactly as if you were using `jest` without having to rewrite all the commands.

toddw:
An api that mimics `npm run` would probably work pretty well `gluestick test -- …args`

PR #775, opened 2017-03-09
Support all Jest CLI options

zamotany:
Solves: #538

### DIFF
--- a/packages/gluestick/src/commands/test/test.js
+++ b/packages/gluestick/src/commands/test/test.js
@@ -65,24 +65,7 @@ const getDebugDefaultConfig = (aliases, webpackRules) => {
   return argv;
 };
 
-const createArgs = (defaultArgs, options) => {
-  const argv = [].concat(defaultArgs);
-  const { coverage, watch, pattern } = options;
-
-  if (coverage) {
-    argv.push('--coverage');
-  }
-  if (watch) {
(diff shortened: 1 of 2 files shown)


packages/gluestick/src/commands/test/test.js
```diff
@@ -62,27 +62,18 @@ const getDebugDefaultConfig = (aliases, webpackRules) => {
...
-    argv.push(pattern);
-  }
-
-  return argv;
+  // Exclude those options to avoid dublication.
+  const optionsToExclude = ['-D', '--debug-test', '-i', '--runInBand', '--watch', '--config', '-c'];
+  return argv.concat(
+    options.filter((option: string): boolean => {
+      return optionsToExclude.findIndex((optionToExclude: string): boolean => {
+        return new RegExp(`^${optionToExclude}.*`).test(option);
```
ferrannp: I would log an info message to the user. Something like "option X is always set by default on debug mode".

ferrannp (dismissed):
If you check https://facebook.github.io/jest/docs/cli.html#running-from-the-command-line, I should also be able to run `npm test -- anyRegex` to run specific tests. For example, I should be able to run: `npm test -- Home`. If I try that now, I will get an error:

```
Cannot read property 'rawArgs' of undefined
```

PR #816, opened 2017-03-16
Extract auto upgrade to separate command

zamotany:
Ref: #811


packages/gluestick/src/commands/autoUpgrade.js
```diff
@@ -0,0 +1,8 @@
+/* @flow */
+import type { Context } from '../types';
+
+const autoUpgrade = require('../autoUpgrade/autoUpgrade');
```
ferrannp: To follow other commands convention, I would move the `autoUpgrade` directory inside the `commands` folder. WDYT?

ferrannp (changes requested):

ferrannp (approved):

PR #829, opened 2017-03-20
Use yarn instead of npm

dratwas:
Solves #827


packages/gluestick-cli/package.json
```diff
@@ -40,6 +40,7 @@
     "fs-extra": "^2.0.0",
     "gluestick-generators": "^1.0.0-rc.8",
     "glob": "7.1.1",
-    "mkdirp": "0.5.1"
+    "mkdirp": "0.5.1",
+    "shelljs": "0.7.6"
```
ferrannp: Do we need a whole new dependency `shelljs` just for using `which`? Is there any other way without adding this dep? If not, I am ok with it.

packages/gluestick-cli/new.js
```diff
@@ -59,16 +60,24 @@ module.exports = (appName, options, exitWithError) => {
...
       stdio: 'inherit',
     },
   );
+  // Remove --npm or -n options cause this is no longer needed
+  const args = commander.rawArgs.slice(2);
+  if (args.indexOf('--npm') >= 0) {
+    args.splice(args.indexOf('--npm'), 1);
+  } else if (args.indexOf('-n') >= 0) {
+    args.splice(args.indexOf('-n'), 1);
+  }
```
ferrannp: This seems unnecessary work to me, isn't it? We can also pass `--dev` or `-d` and you do not remove it. Why do we wanna remove it here in this case?

packages/gluestick-cli/new.js
```diff
@@ -59,16 +60,24 @@ module.exports = (appName, options, exitWithError) => {
   );
 
   spawn.sync(
-    options.yarn ? 'yarn' : 'npm',
+    !options.npm && which('yarn') !== null ? 'yarn' : 'npm',
```
ferrannp: Maybe if the user does not pas `--npm` flag and has no `yarn`, we could print a warning like? `Installing dependencies using npm, consider installing yarn` or something like that.

ferrannp (changes requested):

packages/gluestick-cli/package.json
dratwas: Don't know how i can get only `which` from `shelljs` but there is another library which do the same without rest from `shelljs` 
https://github.com/mathisonian/command-exists

packages/gluestick-cli/new.js
dratwas: Yeah, we don't remove `--dev` or `-d` because we pass it to `gluestick` new command and there we use it. If we don't remove it we get error with message similar to `don't know option --npm/-n`. This is because we  remove option `--yarn/npm` from `gluestick` new command.

ferrannp (approved):

Issue #812, opened 2017-03-16
Improve test coverage


pawlucci:
Missing tests:

- [ ] gluestick-cli
- [ ] gluestick/src/cli/helpers
- [x] gluestick/src/cli/execWithConfig @zamotany
- [ ] gluestick/src/commands/build
- [ ] gluestick/src/commands/dockerize
- [x] gluestick/src/commands/start @zamotany
- [x] gluestick/src/commands/start-client @zamotany
- [ ] gluestick/src/commands/start-server/*
- [x] gluestick/src/commands/test/test @zamotay

Low code coverage:
- [ ] gluestick/src/commands/bin
- [ ] gluestick/src/commands/destroy
- [ ] gluestick/src/plugins/utils
- [ ] gluestick/src/renderer/entry

zamotany:
No longer relevant. If you want to imporve coverage:
1. Go to https://github.com/TrueCar/gluestick/commits/next and from the latest commit click green check icon.
2. Click **Artifacts** tab.
3. Unfold to ` Container 0/home/ubuntu/gluestick/coverage/lcov-report/` and click `index.html`.

PR #831, opened 2017-03-20
Start command test suite

zamotany:
Ref: #812

### DIFF
--- a/packages/gluestick/src/commands/__tests__/start.test.js
+++ b/packages/gluestick/src/commands/__tests__/start.test.js
@@ -9,17 +9,7 @@ const spawnFn = jest.fn(
 );
 jest.setMock('cross-spawn', spawnFn);
 const startCommand = require('../start');
-
-const context = {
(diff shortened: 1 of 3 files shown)


packages/gluestick/src/commands/__tests__/start.test.js
```diff
@@ -0,0 +1,102 @@
+/* @flow */
```
ferrannp, line 1: What is the point of using `flow` if then we `FlowIgnore` everything ? Is because you don't want to write all options that `Context`, `GSConfig`, etc have? Once we write them once, when we can reuse it in every test. Also, can you just use/import `gluestickConfig.js`? I think also @dratwas wrote some mocks, we can reuse them.

Flow is a low level testing tool too. So if we ensure types (not ignoring them), we can avoid some test cases (repetitive).

ferrannp (changes requested):

ferrannp (approved):

Issue #832, opened 2017-03-21
Generating a reducer named "test" causes test failure


marcdel:
# Bug

This is super dumb, but it took me a minute to figure out so I'm opening a ticket for visibility.
I can't add labels, but low priority obvi.

# Reproduction Steps

* `gluestick generate reducer test`
* `gluestick test`

# Current Result

Reducer file is included in the test run.

```
 PASS  src/apps/main/reducers/__tests__/test.test.js
 FAIL  src/apps/main/reducers/test.js
  ● Test suite failed to run

    Your test suite must contain at least one test.

      at onResult (node_modules/jest/node_modules/jest-cli/build/TestRunner.js:192:18
```

# Expected Result

The reducer file `test.js` should not be included in the test run.

PR #833, opened 2017-03-21
Add testRegext to mach test or spec files inside __tests__ dir

zamotany:
Fixes: #832

### DIFF
--- a/packages/gluestick/src/commands/test/test.js
+++ b/packages/gluestick/src/commands/test/test.js
@@ -36,6 +36,7 @@ const getJestDefaultConfig = (aliases: Object, webpackRules: RegExp[]): string[]
   }
 
   const config = {
+    testRegex: '\\/__tests__\\/.*\\.(test|spec)\\.jsx?$',
     moduleNameMapper,
     roots,
     transformIgnorePatterns: [


packages/gluestick/src/commands/test/test.js
```diff
@@ -36,6 +36,7 @@ const getJestDefaultConfig = (aliases: Object, webpackRules: RegExp[]): string[]
   }
 
   const config = {
+    testRegex: '\\/__tests__\\/.*\\.(test|spec)\\.jsx?$',
```
ferrannp, line 39: Isn't this already the [default from Jest](https://facebook.github.io/jest/docs/configuration.html#testregex-string) ?

zamotany: No, the default regex is `(/__tests__/.*|(\\.|/)(test|spec))\\.jsx?$` which will match any file that have `test` or `spec` as a name or every file in `__tests__` folder. This regex will mach files that are inside `__tests__` folder **and** have `test` or `spec` postfix.

ferrannp (approved):

packages/gluestick/src/commands/test/test.js
ferrannp: I see. Very inconvenient to name normal files as `test` but ok.

Issue #809, opened 2017-03-16
Prepublish script to transpile gluestick


pawlucci:
With this, we will be able to get rid of: `require('compiled/gluestick'`) and similars (no need to use `compiled`).

And also, we can remove `transformIgnorePatterns` from `commands/test/test.js`.

We need a wait to still preserve Flow types when transpiling (there are article online about it).

ferrannp:
@pawlucci & @mike866 I added more info about this. Also, I think this should be moved to `Release Candidate` sprint.

ferrannp:
Possibly related to #880.

zamotany:
Closed #990

PR #835, opened 2017-03-22
Prepublish transpilation.

michalchudziak:
Solves #809.

### DIFF
--- a/packages/gluestick/transpile.sh
+++ b/packages/gluestick/transpile.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+transpile () {
+  local dirname=$1
+  local tmpdirname=$dirname"-tmp"
+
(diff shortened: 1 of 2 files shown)


packages/gluestick/transpile.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+transpile () {
+  local dirname=$1
+  local tmpdirname=$dirname"-tmp"
+
+  $(npm bin)/babel $dirname --presets=react,es2015,stage-0 --plugins=transform-decorators-legacy,transform-flow-strip-types -d $tmpdirname && rm -rf $dirname && mv $tmpdirname $dirname & \
```
sethbattin, line 7: Why do we want to transpile in place like this?  This script can only run once before needing to revert the working directory.

michalchudziak: Yeah, true - it was like this only for testing purposes, that's why it's under `wip`.  We are still thinking, about prepublish transpilation and we have mixed feelings. Finally, if we'll decide to go for this solution there should be some other output directory for transpiled code 😇.

zamotany:
Closed in favour of #996

Issue #847, opened 2017-03-23
Create alias for each app apart from "apps"


ferrannp:
It is quite repetitive to do something like `apps/main/components/`, we can do better and have an alias generated for each new app and then do `main/components`, etc. 

Of course depending on the app name you choose, you might create a conflict so I'd be good  we take this into consideration (like error alias already exist).

This is specially important to migrate internal apps .

toddw:
That sounds good but I agree it could be risky. Do you think it would feel as repetitive in day to day app development? I'm wondering if it is just more painful while porting to v1?

I'm okay with making this change if you feel it makes the developer experience better

zamotany:
The only risk that comes to my mind are situation when user will overwrite the default aliases by mistake. @todd do you have any othwr concers about it??

toddw:
overwriting default aliases is a good one @zamotany. We could probably counter that by not allowing an app be named `shared` etc…?

ferrannp:
Yup If you check my original message, I think we should not let them overwrite. So if they try to generate an app that already exists or `shared`, the alias should not be created and we could throw an error. WDYT?

zamotany:
So I made an implementation when it will detect apps based on `src/entries.json` and it will be merged into alias object but the default aliases will overwrite the detected ones.

PR #850, opened 2017-03-23
Auto create aliases for apps

zamotany:
Ref: #847

### DIFF
--- a/packages/gluestick/src/config/webpack/__tests__/getAliasesForApps.test.js
+++ b/packages/gluestick/src/config/webpack/__tests__/getAliasesForApps.test.js
@@ -1,0 +1,101 @@
+/* @flow */
+jest.mock('cwd/noEntries.json', () => ({}), { virtual: true });
+jest.mock('cwd/entryName.json', () => ({
+  '/test1': {
+    component: '',
+  },
+}), { virtual: true });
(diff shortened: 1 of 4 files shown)


packages/gluestick/src/config/webpack/__tests__/getAliasesForApps.test.js
```diff
@@ -0,0 +1,101 @@
...
+    expect(
+      getAliasesForApps(
+        mockGSConfig({
+          entriesPath: 'entries.json',
+          sourcePath: '',
+          appsPath: '',
+        }),
+      ),
+    ).toEqual({
+      Home: 'cwd/src/apps/home',
```
ferrannp, line 107: It took me a while to figure out this test case. I think you can create another `describe` and move the mock there or actually move it here (next to the test) and not in the very top. I am referring to this mock:

```
jest.mock('fs', () => ({
   existsSync: (value) => {
     if (value === 'cwd/test2') {
       return false;
     }
    return true;
   },
 }));
```

You could also add the test case where both path exists and it actually returns 2 aliases.

For the other stuff, LGTM, good job on making all the code to be tested, that's +++++!

ferrannp (changes requested):

Issue #856, opened 2017-03-24
Get Flow to recognize new created aliases


ferrannp:
Related to #847.

PR #858, opened 2017-03-27
Add apps aliases when generating new app

zamotany:
Ref: #856 

Needs to cherrypick commits after merging.


packages/gluestick/generator/predefined/app.js
```diff
@@ -35,7 +35,36 @@ module.exports = (options: GeneratorOptions) => {
         };
         return JSON.stringify(entries, null, '  ');
       },
-    },
+    }, {
+      file: '.flowconfig',
```
ferrannp, line 39: Can we write a test for this logic? So you have a fake `.flowconfig` file, you fake to generated some new apps and compare the previous flow file with the new one ?

ferrannp (changes requested):

ferrannp (approved):

Issue #845, opened 2017-03-23
[V1] Can't start the app in production mode


ferrannp:
**gluestick version 1.0.0-rc.13**

```
gluestick new TestApp
cd TestApp
NODE_ENV=produciton gluestick start
```

It throws:

```
Starting to optimize CSS...
[GlueStick][start-server] Building server entry.
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: spawn /Users/ferrannp/Desktop/TestApp/node_modules/.bin/pm2 ENOENT
    at exports._errnoException (util.js:1012:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:182:32)
    at onErrorNT (internal/child_process.js:348:16)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

PR #860, opened 2017-03-27
Fix pm2 binary resolution and tweak production start

zamotany:
Ref: #845 

Needs to cherrypick commits to next after merging.


ferrannp (commented):
For this to work, do you need to have `pm2` installed globally ?

zamotany:
No pm2 installed locally as it's a dependency of gluestick

ferrannp (approved):

Issue #859, opened 2017-03-27
Support proxies from config/application.js


dratwas:
previous version of gluestick supports proxies from `config/application.js` file.

Check it here:
https://github.com/TrueCar/gluestick/blob/develop/src/lib/server/index.js#L22

We need to implement it in new version of gluestick.

zamotany:
Commits from associated PR must be cherrypicked into `next`.

PR #862, opened 2017-03-27
Set proxies defined in application config

zamotany:
Ref: #859 

Needs to cherrypick commits into `next` after merging.


ferrannp:
@zamotany for cherry-picking to next.

Issue #861, opened 2017-03-27
Apps using gluestick should be able to define/override Jest setup


ferrannp:
In theory, I should be able to define `"jest"` in the app `package.json` and add options like `setupFiles`. 

- Option 1: Our test command parses `package.json` from generated app and options defined by user have precedence
- Option 2: We can provide a file called `jestSetup.js` and our `test` command will look for it if exists. There are some limitations with this approach (`setupFiles` accepts and array) + not handling other options

ferrannp:
This still needs some work. I should be able to provide for example `moduleNameMapper` and that should be merged to ours.

PR #863, opened 2017-03-27
Merge default jest config with custom one from project

zamotany:
Ref: #861 

Commits needs to be cherrypicked into `next` after merging.

### DIFF
--- a/packages/gluestick/src/commands/test/__tests__/test.test.js
+++ b/packages/gluestick/src/commands/test/__tests__/test.test.js
@@ -1,0 +1,82 @@
+/* @flow */
+const spawnMock = jest.fn();
+jest.setMock('cross-spawn', ({ spawn: { sync: spawnMock } }));
+const clone = require('clone');
+const context = require('../../../__tests__/mocks/context');
+const testCommand = require('../test');
+
+const getMockedContext = (aliases, rules) => {
+  const contextConfigCopy = clone(context.config);
+  console.log(contextConfigCopy.webpackConfig.client);
+  contextConfigCopy.webpackConfig.client.resolve.alias = aliases;
+  contextConfigCopy.webpackConfig.client.module.rules = rules;
+  return {
+    ...context,
+    config: contextConfigCopy,
+  };
+};
+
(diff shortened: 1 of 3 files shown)


packages/gluestick/src/commands/test/test.js
```diff
@@ -11,6 +11,22 @@ const JEST_PATH = `${require.resolve('jest').split('jest')[0]}.bin/jest`;
...
+
+  return Object.keys(customConfig).reduce((prev: Object, curr: string): Object => {
+    return {
+      ...prev,
+      [curr]: Array.isArray(customConfig[curr]) && Array.isArray(defaultConfig[curr])
+        ? defaultConfig[curr].concat(customConfig[curr])
+        : customConfig[curr],
+    };
+  }, defaultConfig);
+};
```
ferrannp, line 28: Can we test this logic? It looks to me that `test` command requires finally a test.

ferrannp (changes requested):

packages/gluestick/src/commands/test/__tests__/test.test.js
```diff
@@ -0,0 +1,156 @@
...
+      const jestConfig = JSON.parse(jestMock.run.mock.calls[0][0][1]);
+      expect(jestMock.run.mock.calls[0][0].indexOf('--config')).toBeGreaterThan(-1);
+      Object.keys(jestConfig.moduleNameMapper).forEach((mapper) => {
+        expect(
+          ['woff', 'css', 'alias1', 'alias2'].find((alias) => mapper.includes(alias)),
+        ).not.toBeUndefined();
+      });
+      expect(jestConfig.roots).toEqual(['src', 'test', 'customRoot']);
+      expect(jestConfig.verbose).toBeTruthy();
+      fs.existsSync = originalExistsSync;
```
ferrannp: I'd delete this (and above things related). Jest tests are sandboxed so you don't need here.

You can use `jest.mock(`fs`)`, it will use our mocked `fs.js` and then you just override the methods you want.

ferrannp (approved):

PR #886, opened 2017-04-03
Mock DNS.

michalchudziak:


ferrannp (changes requested):
Can you add a comment above the `node` object saying that this is a temporary fix and should be removed when the apps are able to override this ?

ferrannp (approved):

PR #895, opened 2017-04-03
Add *.server.js functionality back in

toddw:
Prior to GlueStick v1 any files with the extension .server.js would be
mocked in the client bundle but used in the server bundle. This PR adds
that functionality back in.

### DIFF
--- a/packages/gluestick/src/config/webpack/webpack.config.client.js
+++ b/packages/gluestick/src/config/webpack/webpack.config.client.js
@@ -37,7 +37,7 @@ module.exports = (
       { silent: settings.silent, chunk_info_filename: settings.chunk_info_filename },
     ),
     // Make it so *.server.js files return null in client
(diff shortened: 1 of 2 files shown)


packages/gluestick/src/config/webpack/webpack.config.client.js
```diff
@@ -34,6 +36,8 @@ module.exports = (
       deepClone(configuration),
       { silent: settings.silent, chunk_info_filename: settings.chunk_info_filename },
     ),
+    // Make it so *.server.js files return null in client
+    new webpack.NormalModuleReplacementPlugin(/\.server(\.js)?$/, path.join(__dirname, "./mocks/serverFileMock.js")),
```
zamotany: Cay you replace doublequotes with singlequotes??

toddw: updated

PR #898, opened 2017-04-04
Dratwas/support ssr caching

dratwas:
solves #892


zamotany (commented):
I think we need to document this, right?

docs/CachingAndHooks.md
```diff
@@ -16,6 +16,28 @@ Example:
 *Additional optional cache properties:*
 `cacheTTL` - number of seconds to store the cache for a particular route
 
+Gluestick supprots SSR component caching by using [`electrode-react-ssr-caching`](https://github.com/electrode-io/electrode-react-ssr-caching).
```
zamotany: `supprots`? 😄

packages/gluestick/src/renderer/helpers/cacheManager.js
```diff
@@ -17,6 +24,13 @@ const getCacheKey = ({ hostname, url }: { hostname: string, url: string}): strin
 };
 
 module.exports = (logger: Logger, isProduction: boolean): CacheManager => {
+  const enableComponentCaching: EnableComponentCaching = (config) => {
+    if (isProduction && !!config && !!config.components) {
```
zamotany: Are those `!!` really needed?? Both `config` and `config.components` will be evaluated to boolean.

zamotany (changes requested):

packages/gluestick/src/renderer/helpers/cacheManager.js
dratwas: Honestly it's no needed, i was looking at old gluestick :)

zamotany (approved):

Issue #867, opened 2017-03-28
Adding loaders to webpack-additions not working


toddw:
# Bug
additional webpack loaders doesn't seem to be working

## Reproduction steps
1. `gluestick new gs-file-loader`
2. `cd gs-file-loader`
3. Modify `additionalLoaders` in _src/config/webpack-additions.js_ to look like the following:
```
additionalLoaders: [
  {
    test: /\.txt$/,
    use: [
      {
        loader: "file-loader"
      }
    ]
  }
],
```
4. `echo "test 123" > src/apps/main/assets/test.txt`
5. Add the following line to the top of _src/apps/main/components/Home.js_
```
import testFile from "../assets/test.txt";
```
6. `gluestick start`

## Current Result
It generates an error when trying to import _test.txt_ instead of using the file loader.
```
./src/apps/main/assets/test.txt
Module parse failed: /Users/todd/dev/gs-file-loader/src/apps/main/assets/test.txt Unexpected token (1:5)
You may need an appropriate loader to handle this file type.
| test 123
| 
@ ./src/apps/main/components/Home.js 19:12-41
@ ./src/apps/main/containers/HomeApp.js
@ ./src/apps/main/routes.js
@ ./gluestick/clientEntryInit/main.js
@ multi webpack-hot-middleware/client?path=http://localhost:8888/__webpack_hmr webpack/hot/only-dev-server babel-polyfill eventsource-polyfill ./gluestick/clientEntryInit/main.js
```

## Expected Result
It should have used `file-loader` and therefore returned a url to _test.txt_ instead of throwing an error.

zamotany:
You need `gluestick-config-legacy` for that https://github.com/TrueCar/gluestick/tree/next/packages/gluestick-config-legacy

toddw:
Is there a new way to add a loader that doesn't require the legacy plugin?

zamotany:
Not really, we can add that support or make config legacy plugin to be installed by default

toddw:
I thought we had a way to expose the webpack config before building but I don't see that anymore. Was that part of the old "v1"? If that exists still then that would work. Otherwise we should add this as a feature to GlueStick without needing a plugin. It doesn't need to follow the legacy format but there should be a way for people to modify the webpack config

toddw:
original issue: https://github.com/TrueCar/gluestick/issues/407

zamotany:
[...]

PR #902, opened 2017-04-04
Webpack config hooks to modify webpack configs

michalchudziak:
Exposing hooks for overriding webpack config.
Ref: #867


packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
```diff
@@ -14,7 +19,18 @@ const loggerMock = {
...
 describe('config/compileWebpackConfig', () => {
+  beforeAll(() => {
+    process.cwd = () => '.';
+  });
+
+  afterAll(() => {
+    jest.resetAllMocks();
+    process.cwd = originalProcessCwd;
+  });
+
```
ferrannp: i think we need to add more tests cases as we added new functionality.

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
```diff
@@ -14,7 +19,18 @@ const loggerMock = {
...
 
+const originalProcessCwd = process.cwd.bind(process);
+
 describe('config/compileWebpackConfig', () => {
+  beforeAll(() => {
+    process.cwd = () => '.';
+  });
+
+  afterAll(() => {
+    jest.resetAllMocks();
```
ferrannp: I am pretty sure we do not need this line here as we just have one test case (also no need to reset all mocks usually, just clear them).

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
```diff
@@ -14,7 +19,18 @@ const loggerMock = {
   error: jest.fn(),
 };
 
+const originalProcessCwd = process.cwd.bind(process);
```
ferrannp: I'd do `process.cwd.bind({})`.

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
```diff
@@ -1,8 +1,13 @@
-/* @flow */
```
ferrannp: Was Flow intentionally removed? Probably we should re-add it, remove the `FlowIgnore` and reuse config files so flow is ok.

ferrannp (changes requested):

packages/gluestick/generator/templates/gluestick.hooks.js
```diff
@@ -12,5 +12,7 @@ export default {
   postGetCurrentRoute: [],
   postRender: [],
   error: [],
+  webpackClientConfig: (config) => config,
+  webpackServerConfig: (config) => config,
```
dratwas: i think those hooks should be like `: []`

dratwas (changes requested):

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
zamotany: Why?? It makes more sense to bind original context.

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
michalchudziak: Any benefits of adding it if we don't have any typings in this file. We would only need to ignore `process.cwd` bindings.

packages/gluestick/src/config/__tests__/compileWebpackConfig.test.js
ferrannp: To clone it using `bind`, I follow this advice https://github.com/getify/You-Dont-Know-JS/blob/master/this%20%26%20object%20prototypes/ch2.md#safer-this.

Also: http://stackoverflow.com/a/6772648/2277631

packages/gluestick/src/config/compileWebpackConfig.js
```diff
@@ -134,6 +136,18 @@ module.exports = (
         : prev;
     }, serverEnvConfig);
 
+  const pathToWebpackConfigHooks: string = path.join(process.cwd(), 'src', 'webpack.hooks');
```
zamotany: this should be read from gluestick config - `config.GSConfg`

packages/gluestick/src/config/compileWebpackConfig.js
```diff
@@ -134,6 +136,18 @@ module.exports = (
...
+  const { webpackClientConfig, webpackServerConfig }: Hooks =
+    require(pathToWebpackConfigHooks).default;
```
zamotany: what if it doesn't have that file?? I think we should log warn but let it go further

zamotany (changes requested):

packages/gluestick/src/types.js
```diff
@@ -220,6 +220,8 @@ export type Hooks = {
   postGetCurrentRoute?: Hook;
   postRender?: Hook;
   error?: Hook;
+  webpackClientConfig?: Hook;
+  webpackServerConfig?: Hook;
```
dratwas: I think since we have webpack hooks in separated file those types should be also separated.

dratwas: I think since we have webpack hooks in separated file those types should be also separated.

dratwas (changes requested):

packages/gluestick/src/types.js
michalchudziak: Why? They're still same hooks.

dratwas: It'll never exist in this forms. What's the point to adding it here instead of creating new type called webpackHooks or something? Those plugins will not be used together in any case.

zamotany (approved):

dratwas (approved):

ferrannp (approved):

PR #909, opened 2017-04-05
fix console message about gluestick global

bearMountain:
Refine language of gluestick global console message.


zamotany (approved):
Awesome, thanks for correcting it

PR #911, opened 2017-04-06
Change extensions of files, which should be mocked out

### LINKED ISSUE #910: Generated app with 1.2.0 fail to run test command
`Gluestick 1.2.0`.

How to reproduce it:

```
gluestick new App
cd App/
gluestick test
```

Output:
```
Cannot read property 'source' of undefinednpm ERR! Test failed.  See above for more details.
```

Expected: Test to run and pass.

zamotany:
Fixes #910

### DIFF
--- a/packages/gluestick/src/commands/test/test.js
+++ b/packages/gluestick/src/commands/test/test.js
@@ -38,7 +38,7 @@ const getJestDefaultConfig = (aliases: Object, webpackRules: RegExp[]): string[]
 
   // Handling Static Assets = mock them out
   const fileRegex: RegExp = new RegExp(
-    `${webpackRules[3].source}|${webpackRules[4].source}|${webpackRules[5].source}`,
+    `${webpackRules[3].source}|${webpackRules[4].source}`,
   );
   moduleNameMapper[fileRegex.source] = `${TEST_MOCKS_PATH}/fileMock.js`;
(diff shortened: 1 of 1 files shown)


packages/gluestick/src/commands/test/test.js
```diff
@@ -38,7 +38,7 @@ const getJestDefaultConfig = (aliases: Object, webpackRules: RegExp[]): string[]
 
   // Handling Static Assets = mock them out
   const fileRegex: RegExp = new RegExp(
-    `${webpackRules[3].source}|${webpackRules[4].source}|${webpackRules[5].source}`,
+    `${webpackRules[3].source}|${webpackRules[4].source}`,
```
ferrannp: This still can be `undefined` if the user overrides those rules. I'd go with hardcoding this:

```
"moduleNameMapper": {
  "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
  "\\.(css|scss|less)$": "<rootDir>/__mocks__/styleMock.js"
}
```

If then they user adds something else that needs to be handled, they can add that case in `package.json` under `jest`.

ferrannp (changes requested):

ferrannp (approved):

Issue #926, opened 2017-04-12
Fix gluestick predefined generators path resolution and remove project folder if `new` fails


zamotany:
If `gluestick-generator` package is installed inside `node_modules/gluestick/node_modules` it will fail to find predefined generators.

PR #928, opened 2017-04-13
Valid generator paths + remove project folder if `new` fails

zamotany:
Ref #926

### DIFF
--- a/packages/gluestick-cli/new.js
+++ b/packages/gluestick-cli/new.js
@@ -7,6 +7,7 @@ const glob = require('glob');
 const chalk = require('chalk');
 const generate = require('gluestick-generators').default;
 const fetch = require('node-fetch');
+const rimraf = require('rimraf');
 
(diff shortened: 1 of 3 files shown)


packages/gluestick-cli/new.js
```diff
@@ -90,5 +91,15 @@ module.exports = (appName, options, exitWithError) => {
           stdio: 'inherit',
         },
       );
+    })
+    .catch((error) => {
+      rimraf(
+        // Make sure CWD includes appName, we don't want to remove other files
+        process.cwd().includes(appName) ? process.cwd() : path.join(process.cwd(), appName),
```
michalchudziak: Are you sure that it will not result in any undesired behaviour?

zamotany: this is just check if cwd contains appName, because of this line above:
```
process.chdir(appName);
```
so this just ensures it won't delete things we don't want

Issue #955, opened 2017-04-26
Default Babel config doesn't allow for tree shaking / dead code elimination.


threehams:
The default Babel settings for a project don't include `{modules: false}`, so Babel converts ES6 imports to CommonJS. This prevents Webpack from marking unused modules for Uglify to remove (CommonJS can't be statically analyzed).

Whether this is called tree-shaking or dead code elimination (there is so much confusion around these terms), enabling this by default should allow us to bundle only what we need, while still being able to use destructuring with imports.

threehams:
Article from two days ago finally explains this really, really well:
https://medium.freecodecamp.com/tree-shaking-es6-modules-in-webpack-2-1add6672f31b

zamotany:
To get rid of warning add the following line to `src/config/redux-middleware.js`:
```
export const thunkMiddleware = null;
```

PR #959, opened 2017-04-28
Use `modules: false` in babel config by default

zamotany:
Ref: #955


packages/gluestick/src/types.js
```diff
@@ -278,3 +278,8 @@ export type ServerPlugin = {
 
 export type RenderMethod = (root: Object, styleTags: Object[]) =>
 { body: string; head: Object[], additionalScripts?: Object[] };
+
+export type BabelOptions = {
+  plugins: Array<string | any[]>,
```
michalchudziak: I think we could use `Array<string | mixed[]>`

PR #963, opened 2017-05-02
Resolve docker build issues

toddw:
1. .dockerignore was not ignorring node_modules folder when comment lived
next to it

2. gluestick dockerize was running yarn but we were still not
pre-installing it

3. With the speed of yarn and the new folder structure it seemed
reasonable to remove the comment about what we used to do in 0.10.x for
templates/new/package.json

4. We had a random `npm install` command which really wouldn't do
anything without a package.json file existing so removed that

5. Builds were failing for missing node-gyp so we install it globally
now

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -603,9 +603,9 @@ flat-cache@^1.2.1:
     graceful-fs "^4.1.2"
     write "^0.2.1"
 
-flow-bin@^0.38.0:
-  version "0.38.0"
-  resolved "https://registry.yarnpkg.com/flow-bin/-/flow-bin-0.38.0.tgz#3ae096d401c969cc8b5798253fb82381e2d0237a"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -603,9 +603,9 @@ flat-cache@^1.2.1:
     graceful-fs "^4.1.2"
     write "^0.2.1"
 
-flow-bin@^0.38.0:
-  version "0.38.0"
-  resolved "https://registry.yarnpkg.com/flow-bin/-/flow-bin-0.38.0.tgz#3ae096d401c969cc8b5798253fb82381e2d0237a"
+flow-bin@^0.44.2:
```
sethbattin: CircleCI seems to be tripping over this version:
> Wrong version of Flow. The config specifies version ^0.44.2 but this is version 0.44.1

PR #964, opened 2017-05-02
install node-gyp globally with npm install -g instead of yarn global

toddw:
`node-gyp` was still not found and not present in _/usr/local/bin/_ after installing with `yarn global add`. After changing this to use `npm install -g node-gyp` instead it appears in a local test so it should resolve the issue


saltycrane:
👍

Issue #974, opened 2017-05-04
[1.7.5] Static build option generates invalid main.html


marcdel:
# Bug

That static build option seems to generate a server error when rendering which is then output to the html.

# Reproduction Steps

* `git clone https://github.com/marcdel/gs-build-static-error.git`
* `cd gs-build-static-error`
* `gluestick build --static` or `gluestick build --server --static`

# Current Result

main.html contains only the text "Internal Server Error".

```
$ cat build/static/main.html
Internal Server Error%
```

Somewhat related: is there a downside to having the entries html files generated under `build/` rather than `build/static/`? Not a big deal, but having them in a sub-folder means another step in the CI build for a static app rather than just copying the whole build folder over to S3.

# Expected Result

main.html should contain the same html which is generated when running `gluestick start`.

PR #978, opened 2017-05-05
Fix static markup generation

zamotany:
Fixes: #974

### DIFF
--- a/packages/gluestick/src/cli/__tests__/execWithConfig.test.js
+++ b/packages/gluestick/src/cli/__tests__/execWithConfig.test.js
@@ -90,8 +90,7 @@ describe('cli/execWithConfig', () => {
     execWithConfig(command, [{}], {
       skipPlugins: true,
     }, {});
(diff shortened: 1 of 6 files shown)


michalchudziak (approved):

marcdel:
Hey, so `build --static` is no longer printing an exception, but I'm still not entirely sure how to use this option.

Is there a flag or env variable I can pass to specify a relative path to the assets in the generated html file? My use case is deploying a static build to S3 hosting. I need to be able to reference the assets as `./assets/whatever.js` rather than `http://0.0.0.0:8888/assets/whatever.js` as it's currently doing. 

The S3 hosting is currently set up to look for an `index.html` at the root, but I can modify the Jenkins job to move `build/static/main.html` to `build/index.html` before I push the files up to S3, but obviously that doesn't work if the asset references are still `../assets/main.js` or something.

Wondering if you folks have any suggestions for getting around that.

Thanks!

Issue #987, opened 2017-05-08
Restarting dev server - wrong port


dratwas:
If you refresh site during the gluestick is making build you got message `Restarting dev server` with loader below. In console you can see that port which is used by gluestick is wrong. In my case it should be 8888 but is 8880.

PR #994, opened 2017-05-09
Use custom render engine to supply port for poll view

zamotany:
Fixes: #987


packages/gluestick/src/commands/__tests__/start-client.test.js
```diff
@@ -1,4 +1,9 @@
 /* @flow */
+jest.mock('fs', () => ({
```
ferrannp: Maybe some whitespaces in this file ?

ferrannp (changes requested):

Issue #988, opened 2017-05-08
CLI UI revamp


zamotany:
- [x] logger revamp
- [x] header in production/CI/CD
- [x] `new` command
- [x] `build` command
- [x] `start-server` command
- [x] `start-client` command
- [x] `start` command
- [x] `test` command
- [x] `generate` command
- [x] `destroy` command
- [x] `dockerize` command
- [x] `bin` command
- [x] `autoUpgrade` command
- [x] print stats in `build`, `start` and `start-client` commands
- [x] `commandApi test`
- [x] fix Ctrl+C ugly look when running renderer with PM2
- [x] fix progress bar in default mac os terminal app

zamotany:
Preview (WIP):
![gs_start](https://cloud.githubusercontent.com/assets/17573635/25906071/00eb583a-35a4-11e7-820a-c12b32f4d208.gif)

PR #995, opened 2017-05-11
CLU UI revamp

zamotany:
Ref: #988


docs/Commands.md
```diff
@@ -204,3 +204,19 @@ Caching & Hooks documentation is available [here](CachingAndHooks.md).
 ## Styles
 
 Styles documentation is available [here](Styles.md).
+
+## CLI message logging
+
+By default GlueStick will use dark theme when logging messages. If you want to use the light one, either set `GS_LOG_LIGHT` environment variable to `true` or pass `--light` (`-l`) option when running command.
```
ferrannp: I'd write `when running a command.`

docs/Commands.md
```diff
@@ -204,3 +204,19 @@ Caching & Hooks documentation is available [here](CachingAndHooks.md).
...
+
+For better CI/CD and other build tools support, GlueStick will switch to _machine friendly output_ mode, so instead of colorful header prefixing message, you will get log matching the following format:
```
ferrannp: `thus instead of colorful header prefixing message, you will get logs matching the following format`

docs/Commands.md
```diff
@@ -204,3 +204,19 @@ Caching & Hooks documentation is available [here](CachingAndHooks.md).
...
+```
+[GleuStick][<command>][<headerTitle>] <message>
+```
+To switch to _machine friendly_ mode one of the folling condition must be met:
```
ferrannp: add a comma after `mode`?

docs/Commands.md
```diff
@@ -204,3 +204,19 @@ Caching & Hooks documentation is available [here](CachingAndHooks.md).
...
+```
+[GleuStick][<command>][<headerTitle>] <message>
+```
+To switch to _machine friendly_ mode one of the folling condition must be met: 
+
+* `NODE_ENV` must be set to `production`
+* `CI` must be set (to whatever value)
+* `CD` must be set (to whatever value)
+
+However, if you want to overwrite this behaviour pass `GS_LOG_PRETTY=true` when running command.
```
ferrannp: `a command`

packages/gluestick/src/cli/__tests__/logger.test.js
```diff
@@ -4,15 +4,25 @@ const loggerFactory = require('../logger');
 const colorScheme = require('../colorScheme');
 
 const logAndAssert = (message: string, shouldNotLog: string[], loggerInstance: Object) => {
-  Object.keys(loggerInstance).filter(key => key !== 'level').forEach((level: string): void => {
+  ['debug', 'info', 'warn', 'error', 'success'].forEach((level: string): void => {
```
ferrannp: Can we take those values from somewhere? So if we add or remove one we do not need to come back to the test.

packages/gluestick/src/cli/logger.js
```diff
@@ -13,28 +15,78 @@ const levels = {
...
-const loggerFactory = (type: string, level: string) => (...args) => {
+// If you need to find what module prints directly to console.log, uncomment this code
+// and replace `Put some text here` with some word from log you're interested in
+// console._log = console.log.bind(console);
+// console.log = (...args) => {
+//   console._log(...args);
+//   if (args[0].includes('Put some text here')) {
+//     console.trace('here');
+//   }
+// };
```
ferrannp, line 26: This is a bit ugly. Probably some flag when you run `--dev` would be nicer? Although we can do this in the future, not necessary for this change.

ferrannp (changes requested):
Nice and crazy PR!

ferrannp (approved):

Issue #1012, opened 2017-05-24
Accept custom scroll states on route change


mishkinf:
Currently, we are using react-router-scroll in order to support setting the scroll state of the window upon route change. Accept a custom callback to control the scroll state upon window navigation. This will allow us to support things like hash anchor links to deep link to content on a particular page  and leveraging query string params to filter content on the page without having the page pop to the top. 

In Root.js the code might be akin to...

```javascript
const render: Function = applyRouterMiddleware(
  useScroll((prevRouterProps, { location, routes }) => {
    routes.forEach((route) => {
      if (route.useScroll) {
        return route.useScroll(prevRouterProps, location);
      }
    });

    if (routes.some(route => route.ignoreScrollBehavior)) {
      return false;
    }
    return true;
  }),
);
```

and then a client application could define a custom handler for scrolling on each route like so... 

```javascript
// routes definitions
<Route useScroll={customScrollBehavior} path="some_path" component={SomeComponent} />
// end routes definitions

const customScrollBehavior = (prevRouterProps, location) => {
  // first check for hash anchors and if so.. let the browser handle it as normal 
  if (location.hash) { return true; }

  // if the page changes scroll to top of page on path change
  if (prevRouterProps && location.pathname !== prevRouterProps.location.pathname) {
    return [0, 0];
  }

  // default browser scroll behavior
  return true;
};
```

zamotany:
Can you provide some example with `useScroll` prop so it'll be easier to test it?

PR #1015, opened 2017-05-24
Support custom scroll behaviour

zamotany:
Ref: #1012


packages/gluestick/shared/containers/Root.js
```diff
@@ -73,28 +73,39 @@ export default class Root extends Component<DefaultProps, Props, State> {
...
+        // behaviour.
+        const { useScroll: customScrollBehavior } = routes.find(route => (
+          route.useScroll &&
+          prevRouterProps &&
+          prevRouterProps.location.pathname !== location.pathname
+        )) || {};
+
+        if (typeof customScrollBehavior === 'function') {
+          return customScrollBehavior(prevRouterProps, location);
+        }
```
ferrannp: This should be a `function` as part of the API. To ensure this use Flow and/or propTypes. We do not need such checkings in the code (neither tests).

ferrannp (changes requested):

packages/gluestick/shared/containers/Root.js
zamotany: - Can't use `propTypes`, it's Root component but `useScroll` can be defined on Route component.
- Flow types won't cover it - I've checked - I even changed build process to output `Root.js.flow` file to project so it'll get properly check but still flow doesn't detect error.

ferrannp: Still, our API it says it should be a function right? I prefer the code to crash if user passes let's say a boolean that just silently ignore it.

zamotany: Have you seen how helpless are stack traces for some errors? It's better to have that check and add `else` block with meaningful error message - so just by reading it, you would know what you have to fix.

PR #1016, opened 2017-05-24
Add aliases to nameModuleMapper regardless whether it's in custom config

zamotany:
A bug slipped into `1.9.5` which resulted in `aliases` wouldn't have been added, if `nameModuleMapper` was not defined in custom Jest config.

### DIFF
--- a/packages/gluestick/src/commands/test/__tests__/test.test.js
+++ b/packages/gluestick/src/commands/test/__tests__/test.test.js
@@ -16,6 +16,15 @@ jest.mock('cwd/custom/package.json', () => ({
     },
   },
 }), { virtual: true });
+jest.mock('cwd/custom2/package.json', () => ({
+  jest: {
+    roots: ['customRoot'],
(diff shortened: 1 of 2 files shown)


packages/gluestick/src/commands/test/__tests__/test.test.js
```diff
@@ -179,5 +188,46 @@ describe('commands/test/test', () => {
       expect(Object.keys(jestConfig.globals).length).toBe(1);
       expect(jestConfig.globals.__DEV__).toBe(true);
     });
+
+    it('should run jest directly with merged default and custom config (no custom moduleNameMapper)', () => {
```
ferrannp, line 187: I think this is to much copy/paste from the other test. I would write a test case just relevant to the case, so we can define config without custom `moduleNameMapper` and just write a test case that checks that aliases are there nevertheless.

ferrannp (dismissed):

PR #1019, opened 2017-05-24
Updating documentation for custom scroll behavior

mishkinf:
Updating the documentation to give more clarity on the custom scroll behavior


toddw (approved):
:+1:

PR #1020, opened 2017-05-24
Fixing custom scroll behavior to enable custom scroll behavior to be …

mishkinf:
…called even if the current pathname isnt changing to allow for query string param changes and hash params as well.


toddw:
:+1:

PR #1035, opened 2017-06-08
Parallel builds

zamotany:
Ref: ##1033

How to run (npm@5 is not supported):
```
git clone https://github.com/TrueCar/gluestick
cd gluestick
npm run install:npm
cd ../
./gluestick/packages/gluestick-cli/bin/index.js new testApp --npm --dev ./gluestick
cd . testApp
./node_modules/.bin/gluestick build --vendor
./node_modules/.bin/gluestick build --client
```

If vendor DLL bundle is preset it will be used, otherwise it will create normal vendor bundle using `CommonsChunkPlugin`. Also `--skip-if-ok` option in `gluestick build --vendor` will make some checks (https://github.com/TrueCar/gluestick/blob/feature/vendor-dll/packages/gluestick/src/config/vendorDll.js#L47-L112), if every of them fail, it will conclude that vendor dll bundle is fine, and will bail from recompiling it. To build single app: `gluestick build --client -A /<appName>`. To parallely build project:
- build vendor dll `gluestick build --vendor`
- build every app using `gluestick build --client -A /<appName>`
- build server bundle `gluestick build --server`
Now you can `start -P` w/ `NODE_ENV=production` set.

`gluestick-config-legacy` has also been updated to support vendor DLL bundle, so if you're using it, you need to update it `npm i <pathToGluestickRepo>/packages/gluestick-config-legacy`

Documentation can be found [here](https://github.com/TrueCar/gluestick/blob/feature/vendor-dll/docs/Configuration.md#vendoring).

@todd should vendor DLL support also CSS files and loaders?

### DIFF
--- a/packages/gluestick/src/config/vendorDll.js
+++ b/packages/gluestick/src/config/vendorDll.js
@@ -1,13 +1,111 @@
 const path = require('path');
 const fs = require('fs');
 const webpack = require('webpack');
+const sha1 = require('sha1');
 const progressHandler = require('./webpack/progressHandler');
 
 const manifestFilename: string = 'vendor-manifest.json';
 
-const isValid = ({ logger, config }) => {
-  // @TODO
-  return false;
+const getVendorSource = config => {
+  return fs.readFileSync(path.join(process.cwd(), config.GSConfig.vendorSourcePath));
+};
+
+const getDependenciesHash = (config, vendorEntryParts, vendorSource) => {
+  const projectPackage = require(path.join(process.cwd(), 'package.json'));
+  const projectDependencies = {
+    ...projectPackage.devDependencies,
+    ...projectPackage.dependencies,
+  };
+  return sha1(
+    Object.keys(projectDependencies)
+      .filter((name) => {
+        return (
+          vendorEntryParts.find((part) => part.includes(name)) ||
+          vendorSource.includes(name)
+        );
+      })
+      .map((name) => `${name}@${projectDependencies[name]}`)
+      .join('|'),
+  );
+};
+
+const isValid = ({ logger, config }, vendorWebpackConfig) => {
+  const { buildDllPath }: { buildDllPath: string } = config.GSConfig;
+
+  // Check if manifest exists
+  const vendorDllBundleManifestPath = path.join(process.cwd(), buildDllPath, manifestFilename);
+  let manifestContent;
+  try {
+    manifestContent = require(vendorDllBundleManifestPath);
+  } catch (e) {
+    logger.info('Vendor DLL manifest does not exists, recompiling');
+    return false;
+  }
+
+  // Check if DLL bundle itself exists
+  const vendorDllBundlePath = path.join(
+    process.cwd(),
+    config.GSConfig.buildDllPath,
+    `${manifestContent.name.replace('_', '-')}.dll.js`,
+  );
+  if (!fs.existsSync(vendorDllBundlePath)) {
+    logger.info('Vendor DLL bundle does not exists, recompiling');
+    return false;
+  }
+
+  // Check if manifest has validation metadata, if doesn't, it means bundle is not valid, since
+  // we cannot perform other checks
+  if (!manifestContent.validationMetadata) {
+    logger.info('Validation metadata are not defined, recompiling');
+    return false;
+  }
+
+  // Compare elements from vendor entry array stored in manifest against config
+  // which would be used to build DLL bundle
+  const vendorEntryParts: string[] = manifestContent.validationMetadata.entryParts;
+  const currentEntryParts: string[] = vendorWebpackConfig.entry.vendor;
+  if (
+    !Array.isArray(vendorEntryParts) ||
+    vendorEntryParts.length !== currentEntryParts.length ||
+    !vendorEntryParts.every((v, i) => v === currentEntryParts[i]) ||
+    !currentEntryParts.every((v, i) => v === vendorEntryParts[i])
+  ) {
+    logger.info('Vendor DLL entry parts mismatch, recompiling');
+    return false;
+  }
+
+  // Check if vendor source file's hash is the same as stored in manifest
+  const vendorSourceHash: string = manifestContent.validationMetadata.sourceHash;
+  const vendorSource: string = getVendorSource(config);
+  if (vendorSourceHash !== sha1(vendorSource)) {
+    logger.info('Vendor source hash mismatch, recompiling');
+    return false;
+  }
+
+  // Check if dependencies version from project `package.json` has changed
+  const dependenciesHash: string = manifestContent.validationMetadata.dependenciesHash;
+  if (getDependenciesHash(config, vendorEntryParts, vendorSource) !== dependenciesHash) {
+    logger.info('Vendor dependencies hash mismatch, recompiling');
+    return false;
+  }
+
+  logger.success('Vendor DLL bundle is valid, bailing from recompilation');
+  return true;
+};
+
+const injectValidationMetadata = ({ config }, vendorWebpackConfig) => {
+  const { buildDllPath }: { vendorSourcePath: string } = config.GSConfig;
+  const manifestPath = path.join(process.cwd(), buildDllPath, manifestFilename);
+  const manifestContent = JSON.parse(fs.readFileSync(manifestPath));
+  const entryParts = vendorWebpackConfig.entry.vendor;
+  const vendorSource = getVendorSource(config);
+  const dependenciesHash = getDependenciesHash(config, entryParts, vendorSource);
+  manifestContent.validationMetadata = {
+    entryParts,
+    sourceHash: sha1(vendorSource),
+    dependenciesHash,
+  };
+  fs.writeFileSync(manifestPath, JSON.stringify(manifestContent, null, '  '));
 };
 
 const getConfig = ({ logger, config }) => {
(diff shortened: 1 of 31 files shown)


packages/gluestick/src/config/vendorDll.js
```diff
@@ -0,0 +1,208 @@
...
+
+  // Check if manifest exists
+  const vendorDllBundleManifestPath: string = path.join(
+    process.cwd(), buildDllPath, manifestFilename,
+  );
+  let manifestContent: string;
+  try {
+    manifestContent = require(vendorDllBundleManifestPath);
+  } catch (e) {
+    logger.info('Vendor DLL manifest does not exists, recompiling');
```
toddw: should say Vendor DLL manifest does not exist, recompiling

packages/gluestick/src/config/vendorDll.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // Check if DLL bundle itself exists
+  const vendorDllBundlePath: string = path.join(
+    process.cwd(),
+    config.GSConfig.buildDllPath,
+    `${manifestContent.name.replace('_', '-')}.dll.js`,
+  );
+  if (!fs.existsSync(vendorDllBundlePath)) {
+    logger.info('Vendor DLL bundle does not exists, recompiling');
```
toddw: exist

packages/gluestick/src/config/vendorDll.js
```diff
@@ -0,0 +1,208 @@
...
+    return false;
+  }
+
+  // Check if manifest has validation metadata, if doesn't, it means bundle is not valid, since
+  // we cannot perform other checks
+  if (!manifestContent.validationMetadata) {
+    logger.info('Validation metadata are not defined, recompiling');
```
toddw: `Validation metadata is not defined, recompiling`

packages/gluestick/src/config/vendorDll.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // Check if dependencies version from project `package.json` has changed
+  const dependenciesHash: string = manifestContent.validationMetadata.dependenciesHash;
+  if (getDependenciesHash(vendorEntryParts, vendorSource) !== dependenciesHash) {
+    logger.info('Vendor dependencies hash mismatch, recompiling');
+    return false;
+  }
+
+  logger.success('Vendor DLL bundle is valid, bailing from recompilation');
```
toddw: Can we say `skipping recompilation`? `bailing` sounds more like a failure to me

packages/gluestick/src/config/vendorDll.js
```diff
@@ -0,0 +1,208 @@
...
+};
+
+const getConfig = ({ logger, config }: CLIContext, plugins: ConfigPlugin[]): WebpackConfig => {
+  // TODO: check if loaders are necessary, bundle CSS/SASS
+  const appRoot: string = process.cwd();
+  const buildDllPath: string = path.join(process.cwd(), config.GSConfig.buildDllPath);
+  const vendorSourcePath: string = path.join(process.cwd(), config.GSConfig.vendorSourcePath);
+  if (!fs.existsSync(vendorSourcePath)) {
+    logger.fatal(
+      `${vendorSourcePath} does not exists, consider running 'gluestick auto-upgrade' `
```
toddw: `does not exist,`

toddw:
I followed the steps in the PR and I get the following error after running:
```
./gluestick/packages/gluestick-cli/bin/index.js new testApp --npm --dev ./gluestick
```
error:
```
It looks like you've installed the `gluestick` package globally,
but `gluestick` is meant to be installed locally as a dependency.
For proper functionality, please uninstall `gluestick`
and install `gluestick-cli` globally instead.
```

If I ignore that error and run the next command I get the same error again:
Commands:
```
cd . testApp
./node_modules/.bin/gluestick build --vendor
```

output:
```
It looks like you've installed the `gluestick` package globally,
but `gluestick` is meant to be installed locally as a dependency.
For proper functionality, please uninstall `gluestick`
and install `gluestick-cli` globally instead.
```

packages/gluestick/src/config/webpack/webpack.config.client.js
```diff
@@ -50,14 +52,35 @@ module.exports = (
...
+      }),
+    );
+    config.plugins.push(new ChunksPlugin(deepClone(configuration), { appendChunkInfo: true }));
+  } else {
+    logger.info('Vendor DLL bundle not found, using CommonsChunkPlugin');
+    config.plugins.push(
+      new webpack.optimize.CommonsChunkPlugin({
+        name: 'vendor',
+        filename: `vendor${process.env.NODE_ENV === 'production' ? '-[hash]' : ''}.bundle.js`,
+      }),
```
toddw, line 80: This will cause problems in an internal project at TrueCar where we have several white label versions of the same app. Many of the partners have very similar chunks in the partner folder and using CommonsChunkPlugin groups them together. For that reason we can't rely on CommonsChunkPlugin to automatically generate the vendor bundle.

However, digging deeper into the webpack configuration it shows that you can use a `minChunks` function to isolate common chunks to the node_modules folder. Shown here: https://webpack.js.org/plugins/commons-chunk-plugin/#passing-the-minchunks-property-a-function

Lastly, even if we did use CommonsChunkPlugin and isolated it to the node_modules folder I have some concerns that a large app with many sub apps may not benefit as much from this. You could end up with a very large vendor bundle because 2 or 10 of the apps use a large dependency that the other 8 apps do not. Automating this with the CommonsChunkPlugin may work for many use cases but I'm not sure it will work for all use cases. Is it possible to make usage of the CommonsChunkPlugin optional. Instead it would be nice if you could specify a whitelist of vendor modules that will be included in vendor. Including CommonsChunkPlugin should not be included. Really it seems like the handling of how we build out the vendor bundle is a separate task than what we are trying to solve. Parallelizing builds and changing the way we split between apps and the vendor bundle are two separate issues.

toddw:
Can you explain the flow of how the `CommonsChunkPlugin` and `DLL` work? Am I correct in assuming the DLL is created using the CommonsChunkPlugin if the DLL does not exist or is invalid. However, if the DLL exists then the CommonsChunkPlugin is not used because we essentially have the results of that cached in a DLL?

toddw (changes requested):

packages/gluestick/src/config/webpack/webpack.config.client.js
toddw: I think a good solution might be:
We do a very specific `vendor` common chunk using the instructions from the link above. We lock it down to matching _node_modules_. Then we need a way to exclude specific modules from the list. So in the `minChunks` function we match against node_modules and as long as it is not in the blacklist then it will get included.

This way the default behavior is the most common. Most of the time if multiple entry points share the same node_module then they should be extracted into vendor. However, when you come across an edge case where there is a couple of entrypoints that share something the rest of the apps should not, you can blacklist it. For example, a couple of apps might use numeraljs and moment.js which tend to be kind of big libraries, it would be nice to exclude those from the commonschunk if the majority of entry points do not need to use those.

toddw:
We are super close! Overall this looks great, a few minor logging error changes, get it so I can test it, and address the CommonsChunkPlugin challenge and this will be great!

zamotany:
@todd
```
It looks like you've installed the `gluestick` package globally,
but `gluestick` is meant to be installed locally as a dependency.
For proper functionality, please uninstall `gluestick`
and install `gluestick-cli` globally instead.
```
you're using npm@5, which is not supported yet

packages/gluestick/src/config/webpack/webpack.config.client.js
zamotany: I didn't added it, I just moved it, I was like this from the version 1.0, so if it was working before, why to change it?

zamotany:
CommonsChunkPlugin are used as a alternative to DllPlugin, they do not interop. If you have vendor dll bundle created, it will be used **instead of** CommonsChunkPlugin

packages/gluestick/src/config/webpack/webpack.config.client.js
zamotany: Also AFAIK CommonsChunPlugin won't bundle code that is required using `require.ensure` unless `async` is specified. In this case it's not so everything should be fine with partners.

toddw:
The api for `gluestick build --client -A /<appName>` feels weird to me. The forward slash makes me think it would match the _key_ in the entries.json file the url like `build --client -A /my-site` but really it seems to work like `build --client -A /mySite` given the following entry:
```
{
  "/my-site": {
    "name": "mySite",
    "component": "src/apps/my-site/Index.js",
    "routes": "src/apps/my-site/routes.js",
    "reducers": "src/apps/my-site/reducers"
  },
}
```

toddw (approved):

toddw (changes requested):
The api for `gluestick build --client -A /<appName>` feels weird to me. The forward slash makes me think it would match the _key_ in the entries.json file the url like `build --client -A /my-site` but really it seems to work like `build --client -A /mySite` given the following entry:
```
{
  "/my-site": {
    "name": "mySite",
    "component": "src/apps/my-site/Index.js",
    "routes": "src/apps/my-site/routes.js",
    "reducers": "src/apps/my-site/reducers"
  },
}
```

toddw:
should the api be `build --client -A mySite` without `/` instead?

zamotany:
@toddw without `/` gluestick would look for a group, not a single app, it's the same syntax as in `start` command.

michalchudziak (approved):

zamotany:
Merging this PR. `-A` syntax improvement will be done in separate PR.

Issue #1039, opened 2017-06-17
getHttpClient produces errors while removing and setting cookies


sethbattin:
The [`client.interceptors.response.use` function in `getHttpClient`](https://github.com/TrueCar/gluestick/blob/next/packages/gluestick/shared/lib/getHttpClient.js#L71) can produce errors for both setting and removing cookies if headers have already been sent.  Specifically, these errors:

> Error: Can't remove headers after they are sent [line 86](https://github.com/TrueCar/gluestick/blob/next/packages/gluestick/shared/lib/getHttpClient.js#L86)
> Error: Can't set headers after they are sent. [line 88](https://github.com/TrueCar/gluestick/blob/next/packages/gluestick/shared/lib/getHttpClient.js#L88)

In my case, these errors occur when a `gsBeforeRoute` dispatches a GET after receiving a previous GET result.  I will link below to my usage that generates the errors.  With the current release of gluestick-shared, the errors are caught by redux promise middleware and cause silent failures with empty data.   The errors can be caught closer to their source and ignored, which allows the data to be returned normally (PR incoming), but this ignores whatever is the root cause.

My use case is generating email content, which is solely server-side rendering and means I can't use the [workaround in this comment](https://github.com/TrueCar/gluestick/blob/next/packages/gluestick/shared/lib/getHttpClient.js#L81):
> Currently, the suggested solution for dealing with this problem is to make the API requests to A or B in the browser and not in gsBeforeRoute for apps where that is an issue.

Extra info: my `gsBeforeRoute` is sending three simultaneous requests to the same API endpoint.  If I log the cookies that are triggering the errors, I can see that one of them is a session identifier.  That suggests to me that the intent of ["Ensure that any subsequent requests are passing the cookies"](https://github.com/TrueCar/gluestick/blob/next/packages/gluestick/shared/lib/getHttpClient.jss#L91) is not fulfilled.

Related: https://github.com/TrueCar/gluestick/issues/881

sethbattin:
Example error occurs during these requests in a `gsBeforeRoute`:
https://github.com/TrueCar/consumer-frontend/blob/03de1faa6ed7b0374f01dbceed831ec69dcfb43a/src/apps/trade/containers/email/Offer.js#L34

Which follows this request in a `Route.onEnter`:
https://github.com/TrueCar/consumer-frontend/blob/03de1faa6ed7b0374f01dbceed831ec69dcfb43a/src/apps/email/routes.js#L34

[...]

PR #1040, opened 2017-06-17
hotfix #1039 - set cookie after headers sent errors

sethbattin:
This change catches and logs the errors I observed in https://github.com/TrueCar/gluestick/issues/1039  It does not address the root cause of the cookie issues.

See my comment here for the truecar code that exposes this error:  https://github.com/TrueCar/gluestick/issues/1039#issuecomment-309242347


christinebrass:
I spoke with @sethbattin and we determined this was no longer needed.

Issue #1044, opened 2017-06-22
Fail on start, or automatically update, when node_modules is out of date with package.json.


threehams:
Todd and I were talking about this - I spend a lot of time looking at bugs that are fixed in the package.json version of a dependency, but still happen in the stale version in node_modules. This only affects local environments.

On start, can we check the installed packages against `package.json` and either update automatically or (Rails-style) show an error that `npm install` is needed to update before starting?

vdibs:
I'm looking through the code and it looks like this is already implemented (https://github.com/TrueCar/gluestick/tree/next/packages/gluestick/src/commands/autoUpgrade). Is that correct? If so, what needs to be done?

threehams:
Hm.

@toddw What do you think? Gluestick <1 would prompt, and it was... really annoying. A lot of times I'd start up the server, then wondering what was taking so long.

If it failed with an error, but it took multiple seconds, it could have the same issue as the prompt.

Automatically updating could be the most seamless, but would it possibly be a destructive action?

toddw:
@vdibs I believe it checks for dependencies that are out of sync between an app and GlueStick dependencies that are shared like `react`. However, this ticket is talking about the difference between what the package.json file has and what was installed. For example, if my project uses [draft-js](https://www.npmjs.com/package/draft-js) version 0.10.1 and my package.json has:
```
…,
"draft-js": "0.10.1"
…,
```

but the last time I ran `npm install` my package.json file had:
```
…,
"draft-js": "0.9.3"
…,
```

The version of draft-js in the node_modules folder is going to continue to be the 0.9.3 version until I run npm install again, even though my package.json file was updated. The way that I believe we can solve this is to check the node_modules folder to make sure the version specified in package.json is the version installed into the local node_modules folder.

So you would iterate over the dependencies object loaded out of package.json, loop through each dependencies, go into the node_modules/package-name/package.json and look at the version property of that json data. If it does not match then we throw an error suggesting the user run `npm install`.

vdibs:
@toddw  and @threehams - This is the output I have so far for out-of-date, missing, and up-to-date modules. Would either of you prefer different output, info, format, etc.?

<img width="714" alt="screen shot 2017-06-30 at 3 23 49 pm" src="https://user-images.githubusercontent.com/15228252/27755945-237d9b98-5da8-11e7-8ad9-f70055bdb57f.png">

[...]

PR #1051, opened 2017-07-03
Checks for out-of-date dependencies at `gluestick start`

vdibs:
https://github.com/TrueCar/gluestick/issues/1044


packages/gluestick/src/commands/start.js
```diff
@@ -91,6 +96,12 @@ module.exports = (commandApi: CommandAPI, commandArguments: any[]) => {
       );
     }
 
+    try {
+      compareModuleVersions(packagePath, modulePath);
+    } catch (e) {
+      console.log(e);
```
zamotany: Use `logger.warn` or `logger.error` instead of `console.log`.

packages/gluestick/src/commands/start.js
```diff
@@ -91,6 +96,12 @@ module.exports = (commandApi: CommandAPI, commandArguments: any[]) => {
       );
     }
 
+    try {
+      compareModuleVersions(packagePath, modulePath);
```
zamotany: User should be able to disable this check via some option like `--no-dep-check` or sth similar. Also, this whole section should be placed in line 94 before `clientCompilationDonePromise = startClient(`.

packages/gluestick/src/commands/start.js
```diff
@@ -5,6 +5,11 @@ const spawn = require('cross-spawn');
 const startClient = require('./start-client');
 const startServer = require('./start-server');
 const { filterArg } = require('./utils');
+const path = require('path');
+const compareModuleVersions = require('./compareVersions/compareModuleVersions');
+
+const packagePath = require(path.join(process.cwd(), 'package.json'));
```
zamotany: This should be named differently like `packageJson` and placed inside exported function right before the call to  `compareModuleVersions`. Every command thus the exported function is wrapped into try/catch and performs check to ensure that the command is run inside GS project. Leaving this require here in some cases might crash gluestick instead of gracefully failing.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
+
```
zamotany, line 3: Please add `/* @flow */`

packages/gluestick/src/commands/compareVersions/__tests__/compareModuleVersions.test.js
```diff
@@ -0,0 +1,63 @@
+
```
zamotany, line 3: Please add `/* @flow */`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+    const totalNumDependencies = Object.keys(projectPackage.dependencies).length +
+      Object.keys(projectPackage.devDependencies).length;
+    allDependencies = Object.assign(projectPackage.dependencies, projectPackage.devDependencies);
+    if (Object.keys(allDependencies).length < totalNumDependencies) {
+      devDepDifference = true;
+    }
+  }
+
+  Object.keys(allDependencies).forEach(dep => {
+    const tempPath = modulePath.concat('/', dep, packageName);
```
zamotany: Use `path.join(modulePath, dep, packageName)`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+      tempPackage._from.search(fileVersion) === -1)) {
+        console.log(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not satisfy the required version in your package.json')}`);
+        const name = ' '.concat(dep.toString());
+        discrepancies.push(name);
+      }
+    }
+  });
+
+  if (discrepancies.length === 0) {
+    console.log(`${chalk.green('\nNo problems! All node_module versions currently satisfy your package.json.\n')}`);
```
zamotany: Use `logger.success`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+  } else {
+    console.log((devDepDifference) ? `${chalk.red('\nOne or more of these node_modules may have a discrepancy between their dependency and devDependency in package.json.\n')}` : '',
```
zamotany: Use `logger.error`

packages/gluestick/src/commands/start.js
zamotany: Also, pass the `logger` as a third argument

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+ * is either undefined or incorrect.
+ *
+ * Return - returns array of dependencies that are missing or out of date.
+ */
+const compareModuleVersions = (projectPackage, modulePath) => {
+  const discrepancies = [];
+  let devDepDifference = false;
+  let allDependencies = projectPackage;
+
+  if (projectPackage.dependencies === undefined || projectPackage.devDependencies === undefined) {
```
zamotany: Please use short form: `!projectPackage.dependencies || !projectPackage.devDependencies`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+    if (projectPackage.dependencies === undefined || projectPackage.devDependencies === undefined) {
```
zamotany: Same as above

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+      return discrepancies;
+    }
+    allDependencies = projectPackage.dependencies || projectPackage.devDependencies;
+  } else {
+    const totalNumDependencies = Object.keys(projectPackage.dependencies).length +
+      Object.keys(projectPackage.devDependencies).length;
+    allDependencies = Object.assign(projectPackage.dependencies, projectPackage.devDependencies);
```
zamotany: Use spread operator: `allDependencies = { ...projectPackage.dependencies, ...projectPackage.devDependencies }`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  Object.keys(allDependencies).forEach(dep => {
+    const tempPath = modulePath.concat('/', dep, packageName);
+    let tempPackage = true;
+    try {
+      tempPackage = require(tempPath);
+    } catch (e) {
+      console.log(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not exist')}`);
```
zamotany: Use `logger.error`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+      Object.keys(projectPackage.devDependencies).length;
+    allDependencies = Object.assign(projectPackage.dependencies, projectPackage.devDependencies);
+    if (Object.keys(allDependencies).length < totalNumDependencies) {
+      devDepDifference = true;
+    }
+  }
+
+  Object.keys(allDependencies).forEach(dep => {
+    const tempPath = modulePath.concat('/', dep, packageName);
+    let tempPackage = true;
```
zamotany: Flow will cry about this, since it has a boolean assigned and 2 lines below it has a object assigned, so it's better to just leave it as a `undefined` value: `let tempPackage;`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+      tempPackage = require(tempPath);
+    } catch (e) {
+      console.log(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not exist')}`);
+      const name = ' '.concat(dep);
+      discrepancies.push(name);
+      return;
+    }
+
+    if (!tempPackage || !semver.satisfies(tempPackage.version, allDependencies[dep])) {
+      const fileVersion = allDependencies[dep].match(/\d+\.\d+\.\d+/)[0];
```
zamotany: Will throw an error `Cannot read property '0' of null` if depedency is for instance `file:./some-module`.  I suggest to rename the variable to `fileVersionMatch` and remove `[0]` here, then check if `fileVersionMatch` is `null`.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+      if (allDependencies[dep].split(':')[0] !== 'file' || (allDependencies[dep].split(':')[0] === 'file' &&
+      tempPackage._from.search(fileVersion) === -1)) {
+        console.log(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not satisfy the required version in your package.json')}`);
```
zamotany: Use `logger.error`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,71 @@
...
+        const name = ' '.concat(dep.toString());
```
zamotany: What about template literal - ` ${dep.toString()}`?

zamotany (changes requested):
Thanks for the PR, it's getting close to being merged, but it need some polish first.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
vdibs: This is giving me a Syntax Error. Is there anything wrong with using ```Object.assign```?

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
vdibs: The dependencies that are ```file:./some-module``` contain the version in the filename, e.g. ```"react-svg-icons": "file:./npm_private_clone/react-svg-icons-0.8.7.tar.gz"```, so that pattern should always be found. I've tested that case and there hasn't been an issue so far.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
zamotany: No, it's fine, but spread operator is more declarative and more intuitive. Can you specify what error?

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
zamotany: It might work for this case, but it's not like this all the time. For instance I frequently have
```
"gluestick": "../gluestick/packages/gluestick",
```
I use that when testing gluestick on an internal project. So the code above will fail on this. Moreover, it's more common for ppl when using relative/absolute paths not to have any version string in it.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
vdibs: screen shot 2017-07-05 at 11 18 05 am

zamotany: Spread operator is currently a WIP propasal, so it's not supported by node, thus it needs to be transpiled. In order to do that run `npm run build` inside `gluestick` repository every time you made change. This is the easiest solution.

The second, more advanced solution is to run those commands inside the project:
```
npm install --save <path-to-gluestick-repo>/packages/gluestick
gluestick watch
```

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  Object.keys(allDependencies).forEach(dep => {
+    // const tempPath = modulePath.concat('/', dep, packageName);
+    const tempPath = path.join(modulePath, dep, packageName);
+    let tempPackage;
+    try {
+      tempPackage = require(tempPath);
+    } catch (e) {
+      logger.error(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not exist')}`);
```
zamotany: Please remove `chalk.red` from this template string (you can leave `yellow` if you want). `logger.error` will mark the string as a error and we want to maintain consistency between logging various stuff.
So it can look sth like this: `The node_module ${chalk.yellow(dep)} does not exist`.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,72 @@
...
+    const totalNumDependencies = Object.keys(projectPackage.dependencies).length +
+      Object.keys(projectPackage.devDependencies).length;
+    allDependencies = { ...projectPackage.dependencies, ...projectPackage.devDependencies };
+    if (Object.keys(allDependencies).length < totalNumDependencies) {
+      devDepDifference = true;
+    }
+  }
+
+  Object.keys(allDependencies).forEach(dep => {
+    // const tempPath = modulePath.concat('/', dep, packageName);
```
zamotany: I believe it's a leftover, so please remove it.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,72 @@
...
+      logger.error(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not exist')}`);
+      const name = ' '.concat(dep);
+      discrepancies.push(name);
+      return;
+    }
+
+    if (!semver.satisfies(tempPackage.version, allDependencies[dep])) {
+      const fileVersionMatch = allDependencies[dep].match(/\d+\.\d+\.\d+/);
+      if (tempPackage._from.search(fileVersionMatch) === -1) {
+        logger.error(`${chalk.red('The node_module ')}`, `${chalk.yellow(dep)}`, `${chalk.red(' does not satisfy the required version in your package.json')}`);
```
zamotany: Same as above

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,72 @@
...
+        const name = ' '.concat(dep.toString());
+        discrepancies.push(name);
+      }
+    }
+  });
+
+  if (discrepancies.length === 0) {
+    logger.success(`${chalk.green('\nNo problems! All node_module versions currently satisfy your package.json.\n')}`);
```
zamotany: You don't need `chalk.green` here, the message will be pretty even w/o it.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,72 @@
...
+  } else {
+    logger.error((devDepDifference) ? `${chalk.red('\nOne or more of these node_modules may have a discrepancy between their dependency and devDependency in package.json.\n')}` : '',
+    `${chalk.red('\nThe node_module(s)')}`, `${chalk.yellow(discrepancies)}`, `${chalk.red(' is(are) either missing or out of date.\n')}`,
+    `${chalk.red('\nRun')}`, `${chalk.yellow('npm install')}`, `${chalk.red('to resolve this discrepancy.\n')}`);
```
zamotany: Same as above, remove those `chalk.red`

zamotany (changes requested):
Looking pretty good, but there are still some things to sort out.

package.json
```diff
@@ -16,7 +16,10 @@
     "build": "node ./scripts/build.js"
   },
   "dependencies": {
-    "lerna": "2.0.0-rc.5"
+    "babel-jest": "20.0.3",
+    "enzyme": "2.9.1",
+    "lerna": "2.0.0-rc.5",
+    "react": "15.6.1"
```
zamotany: Please remove this leftover. Only `lerna` should be here

packages/gluestick/src/commands/start.js
```diff
@@ -16,6 +20,7 @@ type StartOptions = {
   logLevel?: string;
   runTests: boolean;
   skipBuild: boolean;
+  noDepCheck: boolean;
```
zamotany: This should be `depCheck` - Commander negates options starting with `no`:
- if `--no-dep-check` **is not** passed => `depCheck === true`
- if `--no-dep-check` **is** passed => `depCheck === false`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,73 @@
...
+    } catch (e) {
+      logger.error('The node_module ', `${chalk.yellow(dep)}`, ' does not exist.');
+      const name = ' '.concat(dep);
+      discrepancies.push(name);
+      return;
+    }
+
+    if (allDependencies[dep] && !semver.satisfies(tempPackage.version, allDependencies[dep])) {
+      const fileVersionMatch = allDependencies[dep].match(/\d+\.\d+\.\d+/);
+      if (tempPackage._from.search(fileVersionMatch) === -1) {
```
zamotany: I wouldn't use this - underscore prefixed variables are commonly (this is a non-written rule) used as a private/internal variables and usually are not designed for a external use.

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,73 @@
...
+        logger.error('The node_module ', `${chalk.yellow(dep)}`, ' does not satisfy the required version in your package.json.');
```
zamotany: Use template string

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,73 @@
...
+        const name = ' '.concat(dep.toString());
+        discrepancies.push(name);
+      }
+    }
+  });
+
+  if (discrepancies.length === 0) {
+    logger.success('\nNo problems! All node_module versions currently satisfy your package.json.\n');
```
zamotany: No need for `\n`

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
```diff
@@ -0,0 +1,73 @@
...
+  } else {
+    logger.error((devDepDifference) ? 'One or more of these node_modules may have a discrepancy between their dependency and devDependency in package.json.\n' : '',
```
zamotany: Use template string.

zamotany (changes requested):

packages/gluestick/src/commands/compareVersions/compareModuleVersions.js
vdibs: How risky is it to use `_from`? If one of the underscored variables aren't accessed then there's no way to check those node_modules versions and they'll be skipped.

zamotany (approved):

Issue #722, opened 2017-03-06
Move Jest to parent package (devDependencies)


ferrannp:
The same we did here with `eslint` and `flow` #719 can be done with `jest`. So we install it only once.

The blocker here is that we need to convert `gluestick-shared` tests to Jest (there are very few so it should be fast).

This should also allow us to move presets and plugins needed for `test` env so the problem of running test from parent package should also be solved.

ferrannp:
This will also solve issue of publishing `coverage` to npm since that folder will be created into parent directory. Therefore, we do not need to add `coverage` in every `.npmignore` once this issue is resolved.

ferrannp:
This might need some more thinking, see https://github.com/facebook/jest/issues/3112.

It also needs a solution since running tests form parent package (using `lerna`) does not really work locally.

PR #1052, opened 2017-07-04
Move jest to parent directory

zamotany:
Ref #722


ferrannp (dismissed):
Tests seem not to be running. Output: `echo 'Error: no test specified' "--bail" "--coverage" "-i"`

ferrannp (approved):

PR #1053, opened 2017-07-05
Sets the `dist-tag` for pre-releases

supergibbs:
Sets the `dist-tag` for pre-releases to `alpha`, `beta` or `pre` based on the version. If not a pre-release, will leave as `latest`.

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -16,13 +16,16 @@ const spawnWithErrorHandling = (...args) => {
   }
 };
 
-let tagname = 'latest';
-if (version.includes('alpha')) {
-  tagname = 'alpha';
-} else if (version.includes('beta')) {
-  tagname = 'beta';
-} else if (version.includes('-')) {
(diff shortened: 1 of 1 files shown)


toddw (approved):
:+1:

scripts/deploy.js
```diff
@@ -16,6 +16,15 @@ const spawnWithErrorHandling = (...args) => {
   }
 };
 
+let tagname = 'latest';
+if (version.includes('alpha')) {
```
zamotany: Maybe instead of those cases it's better to use regex for that: `-(.+)\.` so it will match for instance `beta` from `1.13.0-beta.3`, `rc` from `1.0.0-rc.0` and so on.

supergibbs: Good idea! Updated the PR. Will fallback to `pre` if match fails. Relies on pre-release format to be: `<version>-<tag>.<#>`

zamotany (approved):

PR #1064, opened 2017-07-19
Fix cache TTL time unit

zamotany:


zamotany:
Commits in this PR were cherrypicked from #1059 so that it's not a crypto fix.

PR #1065, opened 2017-07-19
Update flow to v0.50.0

supergibbs:
Flow 0.46+ is a lot faster. This updates gluestick to the latest version.

There may be better flow coverage in `start-client.test.js` possible but I fixed the errors and coverage is not any worse.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -18,4 +18,4 @@ suppress_comment=\\(.\\|\n\\)*\\$FlowIgnore
 module.ignore_non_literal_requires=true
 
 [version]
-^0.45.0
+^0.50.0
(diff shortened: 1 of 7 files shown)


zamotany:
Sorry, I just saw that PR after merging #1059 which also updates the flow to `0.50.0`, so I'm going to close this PR. Nevertheless, thanks for putting the time and effort for making this. 💪

supergibbs:
Ah cool, no prob. I like your `jest.fn()` better anyway. You might want to cherry pick https://github.com/TrueCar/gluestick/pull/1065/commits/22c5d5088ba9f64311fcee60e5fc5edb46449387 instead of ignoring kefir in `.flowconfig`

PR #1066, opened 2017-07-19
Update kefir for flow 0.50.0 compatibility

supergibbs:
Just part of the PR https://github.com/TrueCar/gluestick/pull/1065

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,7 +5,6 @@
 .*/node_modules/radium/.*
 .*/node_modules/.*/\(test\).*\.json$
 .*/node_modules/promise/.*
-.*/node_modules/kefir/.*
 
 [include]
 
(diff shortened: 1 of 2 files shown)


supergibbs:
@zamotany Is this right? I noticed in the package update commit you didn't update the yarn.lock file. Should I just update the min version in package.json instead?

zamotany:
`package.json` is a top priority for updating dependencies, so yes.

supergibbs:
Got it but this one actually is a dependency so the lock file is the only way to update it.

zamotany:
I see. This is a dependency of other gluestick depdendency and we cannot really control it apart from `yarn.lock` which only works for `yarn`. So basically the flow will still detect errors if depednencies are installed with npm.

zamotany:
`kefir` is a dependency of `flow-copy-source` so it doesn't really make sense to check it with flow, because even if it fails we don't care.

supergibbs:
Ya, probably doesn't really matter if we're checking it or not. The dependency is set correctly though (`"kefir": "^3.2.0",`) in `flow-copy-source` so npm installs the latest version and works.

supergibbs:
Out of date now and not really needed

PR #1069, opened 2017-07-24
add unzip to list of build packages

yLim92:
Unzip is a very small and very common library needed in many deployment environments. This is needed by TrueCar in our production environment.

Should only add <400 KB (~0.05%) extra size to the docker image.


zamotany (approved):
LGTM 👍

PR #1070, opened 2017-07-25
Fix buildEntires removing client entries thus crashing parallel builds

zamotany:
https://github.com/toddw/gs-parallel-test/issues/1


toddw (approved):
:+1:

PR #1097, opened 2017-09-03
add bash completion

sethbattin:
It's kind-of working:
gluestick_bash_completion

## testing
The unit tests cover the javascript only; they mimic the input format of the bash function.  To functional test the completion against a full project (with multiple entry points), do the following: 

- Check out this branch and override the globally-installed gluestick-cli.
- [Source the completion script and override the location variable.](https://github.com/TrueCar/gluestick/pull/1097/files#diff-733de6a3eeea98143a4fe25bf6f76611R5)
- Update the project's gluestick files, (see [node_modules/gluestick/build/cli.js](https://github.com/TrueCar/gluestick/pull/1097/files#diff-6a60c385a9090ce5fe3dfb036a180ac4) and [this upgrade warning](https://github.com/TrueCar/gluestick/pull/1097/files#diff-497b0fdf7b6939feac1cac1572f6312bR66) )

## The plan so far:

1. Create a bash script to be used as the bash completion entry
2. Call a node script with the full set of completion words
3. Reflect on the commander configuration for global and local gluestick to return completion

## Installation

in `~/.bashrc` or `~/.zsh.rc` or whichever:
`source $(dirname $(readlink -f $(which gluestick)))/completion.sh`
(beware of freebsd problems with readlink)

or, from `/etc/bash_completion.d/`:
`ln -s /usr/bin/node_modules/gluestick/bin/completion.sh gluestick`


packages/gluestick/src/cli/index.js
```diff
@@ -56,18 +56,6 @@ commander.option('-l, --light', 'use light color schema for logging', () => {
 });
 
 commander
-  .command('new')
```
zamotany, line 59: Why `new` command is removed?

packages/gluestick/src/cli/index.js
```diff
@@ -212,4 +200,4 @@ commander.command('*', null, { noHelp: true }).action(cmd => {
   commander.help();
 });
 
-commander.parse(process.argv);
+exports.default = commander;
```
zamotany: You don't have to export it with `default` - `module.exports = commander` is fine.

packages/gluestick/src/cli/index.js
sethbattin: I wasn't 100% certain that it should go, so correct me if I'm wrong about any of this.  The biggest reason I removed it was that it exists in `package/gluestick-cli`.  I guessed that it has been brought along during updates as an oversight instead of deliberately. 

I looked at the commit history to try to figure out whether my guess was right or not.  This one hasn't received all the updates as the cli version, so they're mismatched.  The updates it has received seem to be more cleanup than functional.

That said, I could understand if there's a reason to have both version of the command, so again, tell me if I'm missing it.

zamotany: It is essential to have this command here. The `new` from `gluestick-cli` creates `package.json` and installs dependencies then it spawns `new ` from `gluestick` to create project files.

sethbattin: Ah ok, I definitely should have tried to run `gluestick new`.  Then it would have been obvious I broke it.

I should have mentioned in last comment: this command became obvious when I started iterating over the local commander instance.  I thought it shouldn't be here because it's not part of the cli-api.  No user would ever call this command directly, right?

If so, I might migrate this command elsewhere and update the global's spawned command.  For example, move it to a new command in `packages/gluestick/bin`.  That would allow reflecting on the main commander instance without special filtering, and this file would be readable as a description of the cli interface.  What do you think?

sethbattin: I renamed the command to `_init` and added the no-help commander option to conceal it.  Then I added a filter of the commander settings to exclude commands with a leading '_'.

I tried to test that `gluestick new` was still functional, and I think I succeeded, but it took a lot of linking between projects.  I had to link global gluestick to my working branch (no problem), and then create a project.  I used the `--dev` option, which would fail because it couldn't find the generator template.  Then I linked the project's gluestick dependency to my checkouts `/packages/gluestick`, and installed that and linked it to my checkout's `packages/gluestick-generators`.  After that I was able to execute `gluestick _init` and completion the project creation.

I'm hoping that most of the difficulty vanishes once the change is deployed to npm and all the packages are in the correct location.

zamotany: This might break project generation if user haven't updated `gluestick-cli` and ppl don't do it often. I consider it a breaking change, but I want @toddw opinion about this.

zamotany:
Both `source completion.sh && ch ../src && GS_COMP=$PWD` and `source ../bin/completion.sh && GS_COMP=$PWD` fails locally.

packages/gluestick-cli/src/completion.js
```diff
@@ -0,0 +1,276 @@
...
+    ? command.options.reduce(
+        (opts, curr) => opts.concat([curr.short, curr.long].filter(v => !!v)),
+        [],
+      )
+    : [];
+}
+
+function loadCommanderProject(cwd) {
+  if (!commanderProject) {
+    const completionVersion = '1.14'; // TODO: update me
```
zamotany: What is this version? Should it be `gluestick-cli` or `gluestick`??

zamotany (changes requested):
There are few issues about which I'm concerend:
- Cannot test it locally
- Breaking change with `new` -> `_init`
- https://github.com/TrueCar/gluestick/pull/1097/files#diff-497b0fdf7b6939feac1cac1572f6312bR38
- No official documentation

packages/gluestick-cli/src/completion.js
sethbattin: This checks the project version, so `gluestick`.  To be honest I don't know how the publishing process works - I guess I assumed that one release of the project updated all packages to the same version.  Do I have that wrong?

packages/gluestick/src/cli/index.js
sethbattin: That makes sense.  I had imagined that scenario to be impossible because a user couldn't be running the completion without updating `gluestick-cli`, but this change would affect local gluestick without forcing a `-cli` update.  So failure scenario is:
1. previous version of `gluestick-cli`, command `gluestick new` installs updated version of `gluestick`
2. Updated package does not provides the `_init` command, not `new`.
3. Previous `-cli` fails attempting to spawn `new`, so the project generator never runs.

I don't see any way to avoid backing out this change.  I tried to look at https://github.com/TrueCar/gluestick/pull/1097/files#diff-01fb1ca83ea3c067c9cc6c70105fb1f7R102 to see if there was some recourse there based on how local `new` is called, but that would still require an updated `gluestick-cli`.  I can still add `noHelp`, and I make my commander inspection filter be explicit.

Out of curiosity: this problem won't be solved by semantic versioning.  Users could always have stale `gluestick-cli`, even after we publish 2.0.  What's the strategy we will apply when that happens?  And...any chance we could apply it right now?  :)

sethbattin:
@zamotany regarding the failure of the bash file - can you elaborate about how it failed?  I know Todd bumped into the `readlink` problem.  If that's going to happen to every OSX machine then I should just replace it.

About the docs: what do you feel would be sufficient?  I do plan to record the installation method (probably mimicking https://docs.npmjs.com/cli/completion).

packages/gluestick/src/cli/index.js
toddw: Is there no way to make it so tab completion simply does nothing in the old version but does work in the new version? Does this have to be a breaking change?

sethbattin: No, totally not required.  I can back it out.  I thought I had a thing that should be changed, but the breaking isn't worth the benefit.

packages/gluestick-cli/src/completion.js
zamotany: I don't really know which version you want, but if you need the latest version from npm then you can do sth similar to: https://github.com/TrueCar/gluestick/blob/next/packages/gluestick-cli/src/new.js#L12-L17

but if you need installed version of gluestick from project then you can find it here
https://github.com/TrueCar/gluestick/blob/next/packages/gluestick-cli/src/cli.js#L24-L35

zamotany:
@sethbattin Here's the error I get:

```bash
$ source completion.sh && ch ../src && GS_COMP=$PWD
readlink: illegal option -- f
usage: readlink [-n] [file ...]
usage: dirname path
bash: ch: command not found
```

About documentation: just some steps how to install completion.
I was also thinking about having a command in `gluestick-cli` that will setup auto completion automatically for you, something like `gluestick install-completion` or similar.

packages/gluestick-cli/src/completion.js
sethbattin: Oh, I misunderstood - yes I check the locally installed version below, and this value is compared against it.  It needs to be the version in which these changes are released.

The purpose is to avoid trying to import the local commander file when it contains `commander.parse`.  That could happen with and updated version of `gluestick-cli` trying to import in a project with an older local package dependency.

sethbattin:
@zamotany yes, that's the same problem we saw.  Apparently `readlink -f` it is documented for openBSD, but OSX's version mismatches that.  Todd had to explicitly install some extra gnu utilities to make it work.  That's clearly an unreasonable expectation, so it needs to go.

I think the `completion-install` idea can replace it anyway.  If it works exactly like `npm completion`, then it just outputs the contents of the shell script, but it can detect the `gluestick-cli` bin directory when it runs.  Then no need for `readlink` or the `$GS_COMP` export.

And extra benefit: it doesn't have to work in any environment, because it can be piped to whatever shell config the user prefers
`$ gluestick completion >> ~/.bashrc`
or
`# gluestick completion > /etc/bash_completion.d/gluestick`

sethbattin:
@zamotany @toddw I reverted the change to the local new command, and I implemented the `npm-completion` style command to output the completion shell script.  After installing this version of gluestick-cli, you may now run this to install it:
```
gluestick completion > /etc/bash_completion.d/gluestick
```
I also updated the version check to 1.15, assuming that's going to be the version under which this feature is released.

zamotany:
Not sure if this `you_should_update_the_gluestick_dependency_to_at_least_1.15` should be printed.
screen shot 2017-10-13 at 12 49 39

zamotany (approved):
LGTM

sethbattin:
@zamotany I could be convinced to drop it; it's there because I couldn't think of another place to output a completion warning message.  It does mean exactly what it says, also see [this comment](https://github.com/TrueCar/gluestick/pull/1097/files#diff-497b0fdf7b6939feac1cac1572f6312bR50).  Once we get projects updated, that message goes away and the completion will start using the commander instance.

I am hoping that we can roll a release out and upgrade most of our projects, then later on inform folks of the new CLI feature.  That way we we'll have everything working as intended before folks try to start using it.

PR #1098, opened 2017-09-08
[SELA-3087]: Add new option to build command

lukewalczak:
https://jira.true.sh/browse/SELA-3087


zamotany (changes requested):
Code LGMT 👍, fix flow and we will merge this.

zamotany (approved):

PR #1099, opened 2017-09-11
Support for Regexes as route paths

zamotany:


packages/gluestick/src/renderer/helpers/parseRoutePath.js
```diff
@@ -0,0 +1,24 @@
...
+
+type PathRegexp = RegExp & {
+  fromRegex?: boolean,
+  keys?: string[],
+};
+
+module.exports = function parseRoutePath(routePath: string): PathRegexp {
+  const match = /\/(.+)\/([gmiyu]*)$/.exec(routePath);
+  if (match) {
+    const [, pattern, modifiers] = match;
```
ferrannp, line 13: Is that `,` wrong?

ferrannp (changes requested):

packages/gluestick/src/renderer/helpers/parseRoutePath.js
ferrannp: ok like `_,` ...

ferrannp (approved):

PR #1102, opened 2017-09-12
[SELA-3273]: Name function for profiling app

lukewalczak:
https://jira.true.sh/browse/SELA-3273

### DIFF
--- a/packages/gluestick/src/renderer/helpers/getRequirementsFromEntry.js
+++ b/packages/gluestick/src/renderer/helpers/getRequirementsFromEntry.js
@@ -31,11 +31,11 @@ const getSortedEntries = (entries: Entries): string[] => {
  * variables that the server needs to render. These variables include Index,
  * store, getRoutes and fileName.
  */
-module.exports = (
+module.exports = function renderRequirements(
   { logger }: Context,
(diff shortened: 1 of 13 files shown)


packages/gluestick/src/renderer/helpers/getRequirementsFromEntry.js
```diff
@@ -31,11 +31,11 @@ const getSortedEntries = (entries: Entries): string[] => {
  * variables that the server needs to render. These variables include Index,
  * store, getRoutes and fileName.
  */
-module.exports = (
+module.exports = function renderRequirements(
```
zamotany: Should be named `getRenderRequirements`

packages/gluestick/src/renderer/render.js
```diff
@@ -37,7 +37,7 @@ type AssetsCacheOpts = {
   cacheManager: Object,
 };
 
-module.exports = (
+module.exports = function renderOutput(
```
zamotany: `render` should be enought

zamotany (changes requested):

zamotany (approved):

PR #1107, opened 2017-09-15
resolve regex support bug

toddw:
Non regex entrypoint keys were converted to a regex that required the
path to end on the key. It did not support routes that begin with that
key. This update takes advantage of path-to-regex's options and adds
`end:false` so that it will match on routes that begin with the matched
key.


christinebrass (approved):
👍

christinebrass (approved):

PR #1109, opened 2017-09-18
Alias react and react-dom to dist version in server config

zamotany:


packages/gluestick/src/config/webpack/webpack.config.server.prod.js
```diff
@@ -4,5 +4,14 @@ import type { WebpackConfig } from '../../types';
 
 module.exports = (serverConfig: WebpackConfig): WebpackConfig => ({
   ...serverConfig,
+  resolve: {
+    ...serverConfig.resolve,
+    alias: {
+      ...(serverConfig.resolve ? serverConfig.resolve.alias : {}),
+      react: require.resolve('react/dist/react.js'),
+      'react-dom/server': require.resolve('react-dom/dist/react-dom-server.js'),
+      'react-dom': require.resolve('react-dom/dist/react-dom.js'),
```
ferrannp, line 13: Shall we allow them to override those values if they want ?

ferrannp (changes requested):

packages/gluestick/src/config/webpack/webpack.config.server.prod.js
zamotany: They can still overwrite it in `webpack.hooks.js`

ferrannp (approved):

PR #1110, opened 2017-09-18
SELA-3277: GlueStick: gluestick reset-hard fails with npm5

ferrannp:


packages/gluestick-cli/src/reset.js
```diff
@@ -1,8 +1,18 @@
 const { execSync } = require('child_process');
 
+const _cleanCache = () => {
+  const npmVersion = execSync('npm -v').toString().trim();
+  const majorVersion = npmVersion.split('.')[0];
+  // As of npm@5, the npm cache self-heals from corruption issues and data extracted from the cache
```
zamotany: The point of `reset-hard` is not to heal the cache but to remvoe gluestick pakages from cache. This is important in development, because if gluestick is installed it will be stored in cache and next installs will use that cache, which is not desired behaviour if in the meantime gluestick's source code has changed / got updated.

zamotany (changes requested):

packages/gluestick-cli/src/reset.js
ferrannp: So you suggest just to use `--force` ? Reading the warnings, it should not be necessary...

zamotany: Yes.

Warning are not talking about our case and are irrelevant for us.

zamotany (approved):

PR #1111, opened 2017-09-18
SELA-3097: Remove `yarn` message from `gluestick new` when `yarn` is not installed

ferrannp:


packages/gluestick-cli/src/new.js
```diff
@@ -77,13 +76,6 @@ module.exports = (appName, options, exitWithError) => {
         });
 
         const isYarnAvailable = !spawn.sync('yarn', ['-V']).error;
```
zamotany: I think we should replace 'V' with 'v' also.

zamotany (approved):

zamotany (approved):

PR #1128, opened 2017-10-12
Migrate off React.PropTypes to prop-types package

zamotany:


packages/gluestick/src/renderer/helpers/cacheManager.js
```diff
@@ -9,7 +9,7 @@ import type {
 } from '../../types';
 
 const LRU = require('lru-cache');
-const SSRCaching = require('electrode-react-ssr-caching');
+// const SSRCaching = require('electrode-react-ssr-caching');
```
ferrannp: Forgot some commented code in this file.

ferrannp (changes requested):

ferrannp (approved):

PR #1139, opened 2017-10-23
Gracefully handle bad query parameters

hhtran:
This PR is to fix an issue where bad query params will make react-router to throw an error and cause pages to throw a 500 instead of gracefully handling the bad query param.

The source of this bug is query-string throwing an error inside of react-router, so this PR tries to gracefully handle it in a try/catch block, interpreting the route without the query params altogether if the queries are bad

Bug discussion can be seen here for react-router v3:
https://github.com/ReactTraining/history/issues/312

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,7 @@
     "gitbook-plugin-github": "2.0.0",
     "gitbook-plugin-prism": "^2.3.0",
     "jest": "20.0.4",
-    "lerna": "2.0.0-rc.5",
-    "query-string": "^4.2.2"
+    "lerna": "2.0.0-rc.5"
   },
   "devDependencies": {
     "babel-core": "^6.24.1",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "gitbook-plugin-prism": "^2.3.0",
     "history": "^3.0.0",
     "jest": "20.0.4",
-    "lerna": "2.0.0-rc.5"
+    "lerna": "2.0.0-rc.5",
+    "query-string": "^4.2.2"
   },
   "devDependencies": {
     "babel-core": "^6.24.1",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "gitbook-plugin-edit-link": "^2.0.2",
     "gitbook-plugin-github": "2.0.0",
     "gitbook-plugin-prism": "^2.3.0",
+    "history": "^3.0.0",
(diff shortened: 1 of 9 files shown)


hhtran:
[clabot:check]

toddw:
[clabot:check]

package.json
```diff
@@ -28,7 +28,8 @@
     "gitbook-plugin-github": "2.0.0",
     "gitbook-plugin-prism": "^2.3.0",
     "jest": "20.0.4",
-    "lerna": "2.0.0-rc.5"
+    "lerna": "2.0.0-rc.5",
+    "query-string": "^4.2.2"
```
zamotany: Since this is a monorepo, this package should be installed inside `packages/gluestick`. This `package.json` is for dependencies used to develop gluestick packages, each package have it's own `package.json` where the actual dependencies should be specified.

zamotany (changes requested):

packages/gluestick/package.json
```diff
@@ -102,6 +102,7 @@
     "oy-vey": "0.9.0",
     "pm2": "2.3.0",
     "progress": "2.0.0",
+    "query-string": "^4.2.2",
```
hhtran, line 105: This doesn't affect our yarn.lock file because this is simply explicitly adding the query-string package that history uses

package.json
hhtran: Got it. Good catch! Moved this dependency to the gluestick package instead

hhtran:
@zamotany are there any other things that I should address in this PR?

packages/gluestick/src/renderer/helpers/matchRoute.js
```diff
@@ -15,8 +17,21 @@ module.exports = function matchRoute(
...
+    // React Router v3 does not catch the exceptions thrown by query-string,
+    // which causes 500 errors on pages to be shown
+    // We are handling it gracefully by omitting the query instead
+    let location: string;
+    try {
+      queryString.parse(req.url);
+      location = req.url;
+    } catch (e) {
+      const parsedUrl: Object = parseUrl(req.url);
+      location = parsedUrl && parsedUrl.pathname ? parsedUrl.pathname : '';
```
hbstone: `location = parsedUrl && parsedUrl.pathname || '';` should do the same, and avoids the redundancy of the ternary in this case.

zamotany (approved):

PR #1143, opened 2017-10-31
Load polyfills on demand

zamotany:


toddw (commented):
I know that I said the polyfill service would be okay to use but as I see the code… Blocking our all of our app code based on a free 3rd party service doesn't seem like a wise choice. Is there any way we can accomplish this without a 3rd party requirement?

toddw:
https://philipwalton.com/articles/loading-polyfills-only-when-needed/ documents an approach really well that I think would be better suited

packages/gluestick/src/generator/templates/Index.js
```diff
@@ -29,7 +29,7 @@ export default class Index extends Component {
 
     const helmet = Helmet.rewind();
     const bodyAttributes = BodyAttributes.rewind();
-
+    const features = ["fetch", "Promise"];
```
threehams: fetch require Promise, so we only need to check for fetch.

packages/gluestick/src/generator/templates/Index.js
```diff
@@ -43,6 +43,11 @@ export default class Index extends Component {
         </head>
         <body {...bodyAttributes}>
           {body /* DO NOT REMOVE */}
+          <script
+            type="text/javascript"
+            // eslint-disable-next-line react/no-danger
+            dangerouslySetInnerHTML={{ __html: \`var features = [\${features.map(e => \`'\${e}'\`).join()}]\` }}
```
threehams: Does this really need to be configurable? At some point we need to draw the line on convention vs. configuration, since this requires a single-use global variable, adds a level of indirection, and makes this a breaking change (the change to Index.js means that all consumers will see very, very confusing errors on upgrade).

`Uncaught ReferenceError: features is not defined`

Even if we make it configurable, "features" is a dangerous global variable to set, and is going to cause someone a _lot_ of debugging pain when they accidentally reference `window.features`.

packages/gluestick/src/renderer/helpers/getAssetsLoader.js
```diff
@@ -13,12 +13,13 @@ module.exports = function getAssetsLoader(
   loadjsConfig: LoadJSConfig,
   entryPointBundle: string,
   vendorBundle: string,
+  polyfillBundle?: string = 'https://cdn.polyfill.io/v2/polyfill.min.js',
```
threehams: Why is polyfillBundle set up as an optional property with a default? Why not just set it as a constant inside this file? This is a function that's internal to Gluestick - expanding its API with an optional and defaulted argument is just confusing.

packages/gluestick/src/renderer/helpers/getAssetsLoader.js
```diff
@@ -27,8 +28,28 @@ module.exports = function getAssetsLoader(
       });
     };
 
+    function isPolyfillNeeded() {
+      var needed = false;
+      while (!needed && features.length) {
+        var feature = features.shift();
```
threehams: `for` would be a lot less confusing. This is mutating the global "features" variable as it reads it. Someone trying to figure out what it's doing after the fact would just see an empty array.

threehams (changes requested):
Please test this in a real app, using a browser that doesn't support fetch (like IE11). On local testing, the bundle size is larger than before this change, and IE11 fails with a minified error. The polyfill script tag is set to async.

zamotany:
@threehams 
With `babel-polyfill`:
screen shot 2017-11-02 at 12 32 16

With `regenerator-runtime` only:
screen shot 2017-11-02 at 12 30 08

For new app, build in production.

toddw (changes requested):
This is still blocking our JS relying on a free 3rd party service. Please see my comments above and read the blog post for an alternative approach

zamotany:
This is the approach from the article. Check for unsupported features and load polyfill only if needed.

zamotany:
Just tested it on IE11, and it works after adding `<default_gateway> 0.0.0.0` to `hosts` file, since `0.0.0.0` is a default `publicPath` in development.

packages/gluestick/src/renderer/helpers/getAssetsLoader.js
```diff
@@ -18,7 +18,7 @@ module.exports = function getAssetsLoader(
...
 
-    var loadVendorThenEntry = function() {
+    function loadVendorThenEntry() {
       loadjs(['${vendorBundle}', '${entryPointBundle}'], {
```
toddw: As we discussed, we should load the polyfills as part of this array since it has `async: false`. This way they load in parallel

toddw (changes requested):
load in parallel

packages/gluestick/src/renderer/helpers/getAssetsLoader.js
```diff
@@ -27,8 +27,21 @@ module.exports = function getAssetsLoader(
...
     };
 
+    function isPolyfillNeeded() {
+      return !window['fetch'];
+    }
+
     document.addEventListener('DOMContentLoaded', function() {
-      loadVendorThenEntry();
+      if (isPolyfillNeeded()) {
+        loadjs('https://cdn.polyfill.io/v2/polyfill.min.js', {
```
toddw: Is there any way we could host the polyfill through GlueStick… I'm wondering what happens if polyfill.io goes down

toddw: As we discussed, if we served this file directly from GlueStick, this might work: https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.26.0/polyfill.min.js

packages/gluestick/src/renderer/helpers/getAssetsLoader.js
```diff
@@ -13,22 +13,34 @@ module.exports = function getAssetsLoader(
   loadjsConfig: LoadJSConfig,
   entryPointBundle: string,
   vendorBundle: string,
+  config: Object,
 ): string {
+  const { host, protocol, ports } = config.GSConfig;
+  const hostAddress =
+    process.env.NODE_ENV === 'production'
+      ? ''
+      : `${protocol}://${host}:${ports.server}`;
```
toddw: Is this reliable? In production we want assets to be served from the CDN location where all of the other assets would be served. Can we not use webpack file loader to url for polyfill.min.js?

Something like:
```
import polyfillUrl from "file-loader!./polyfill.min.js"
```

toddw: Or better yet, couldn't we use the same strategy as we did here: https://github.com/TrueCar/gluestick/blob/232d039b90dc904c99fde29f5cdda5789d60795a/packages/gluestick/src/renderer/helpers/linkAssets.js#L65 with `getAssetPathForFile`

PR #1146, opened 2017-11-07
withDataLoader HOC

zamotany:


packages/gluestick/shared/lib/route-helper.js
```diff
@@ -5,6 +5,16 @@ import type { GetBeforeRoute } from '../types';
...
   const c: Object = component.WrappedComponent || component;
+
+  if (c.onEnter) {
+    return typeof window === 'undefined'
+      ? c.onEnter
+      : (...args) => {
+          c.onEnter(...args);
+          return Promise.resolve();
+        };
+  }
```
toddw: I'm having trouble understanding this code. Can you explain this? I'm not sure why we aren't returning `return c.onEnter(...args)` but that is also because I don't entirely know what is happening.

zamotany: We do not return `c.onEnter(...args)` in browser to make transition between pages async/non-blocking, but to block rendering on server. Once we upgrade to RR4 this part will be gone and moved into `componentDidMount`.

toddw (changes requested):
I tried building a sample application with this branch and server side rendering does not appear to be working. Can you please check out the code in the sample app and let me know if I'm doing something wrong. I would consider this to be a blocker for this feature. https://github.com/toddw/gluestick-test-with-dataloader

packages/gluestick/shared/lib/route-helper.js
```diff
@@ -13,6 +23,13 @@ const getBeforeRoute: GetBeforeRoute = (component = {}) => {
     );
   }
 
+  if (c.fetchData) {
```
toddw: should this be `c.gsBeforeRoute`?

packages/gluestick/shared/lib/route-helper.js
```diff
@@ -5,6 +5,16 @@ import type { GetBeforeRoute } from '../types';
...
+      ? c.onEnter
+      : (...args) => {
+          c.onEnter(...args);
+          return Promise.resolve();
+        };
+  }
+
   // @deprecated since 0.0.1
   // check for deprecated fetchData method
   if (c.fetchData) {
```
toddw: By this point we should remove support for the deprecated `.fetchData`

toddw (approved):

toddw (changes requested):
With the current code it seems as though `gsBeforeRoute` will silently be ignored if `onEnter` is defined inside of `withDataLoader`. Since this is the beginning of deprecating `gsBeforeRoute` it seems like `gsBeforeRoute` should still work with the deprecation warning even if both are defined.

toddw (approved):
Looks good to me! This seems as though it would merge into v2 since we are deprecating `gsBeforeRoute`

PR #1147, opened 2017-11-11
Resolve 3rd party cookie security issue

toddw:
Resolves https://github.com/TrueCar/gluestick-shared/issues/21

- [x] Update getHttpClient
- [x] test manually
- [x] update unit tests


packages/gluestick/shared/lib/getHttpClient.js
```diff
@@ -122,3 +130,13 @@ export default function getHttpClient(
 
   return client;
 }
+
+function hostMatch(config) {
+  if (!config || !config.baseURL || !config.url) {
+    return false;
+  }
+
+  const originalHostName = new URL(config.baseURL).hostname;
```
zamotany: Please use `parse` function from `url` module, since there's no `URL` constructor in node 6. This is also the reason why CI is failing.

packages/gluestick/shared/lib/getHttpClient.js
```diff
@@ -122,3 +130,13 @@ export default function getHttpClient(
...
+  const outgoingRequestHostName = new URL(config.url).hostname;
```
zamotany: As above

zamotany (changes requested):

zamotany (approved):

PR #1150, opened 2017-11-14
Add auto-upgrade to postinstall script

zamotany:


toddw (dismissed):
I tried `gluestick new appName --dev ./gluestick` while this branch was checked out and I'm not getting any of this functionality. Am I missing a step? It doesn't look like this is part of `gluestick-cli` so I don't expect I need to run that right?

toddw (approved):
LGTM!

PR #1166, opened 2018-03-06
Newly generated apps should run on enzyme 3

christinebrass:
This is a patch to the GlueStick 2.0.0 pre-release. All newly generated apps should run on Enzyme 3. This PR upgrades that dependency for new apps  .


toddw (approved):
:+1: looks good!

PR #1168, opened 2018-03-23
Apply patches to properly load bunyan config file

christinebrass:
The Bunyan plugin was not actually loading in an app's config file, therefore always using a default Bunyan configuration.

Changes made:
* Use requireModule to digest bunyan config file
* Update test to mock requireModule
* Add test for empty config case

### DIFF
--- a/packages/gluestick-plugin-bunyan/src/server.js
+++ b/packages/gluestick-plugin-bunyan/src/server.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import bunyan from 'bunyan';
+import path from 'path';
 
 export type Options = {
   name: string,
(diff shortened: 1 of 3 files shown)


packages/gluestick-plugin-bunyan/src/server.js
```diff
@@ -19,24 +20,41 @@ export type Logger = {
...
   name: 'default name',
 };
 
-const bunyanPlugin = (): { logger: Logger } => {
-  let options: Options = defaultSettings;
+const bunyanPlugin = (
+  opts: Object,
+  { requireModule, defaultLogger }: PluginUtilities,
+): { logger?: Logger } => {
+  let options: Options = { ...defaultSettings };
```
threehams: Is there any reason to make a shallow copy here? Nothing in this file is changing it.

packages/gluestick-plugin-bunyan/src/server.js
```diff
@@ -19,24 +20,41 @@ export type Logger = {
...
+};
+
 const defaultSettings: Options = {
   name: 'default name',
 };
 
-const bunyanPlugin = (): { logger: Logger } => {
-  let options: Options = defaultSettings;
+const bunyanPlugin = (
+  opts: Object,
```
threehams: This seems to be unused - is it just here by convention? Maybe it could be typed to `null`.

packages/gluestick-plugin-bunyan/src/server.js
```diff
@@ -19,24 +20,41 @@ export type Logger = {
...
+  { requireModule, defaultLogger }: PluginUtilities,
+): { logger?: Logger } => {
+  let options: Options = { ...defaultSettings };
   try {
-    // $FlowIgnore
-    const userSettings: Options = require('src/bunyan.config.js').default;
+    const userSettings: Options = requireModule(
+      path.join(process.cwd(), 'src/bunyan.config.js'),
```
threehams: I've gotten to like the aliases because of the automatic typing it gives us, though there's no reason to change this right now.

packages/gluestick-plugin-bunyan/src/server.js
```diff
@@ -19,19 +20,29 @@ export type Logger = {
...
 };
 
-const bunyanPlugin = (): { logger: Logger } => {
-  let options: Options = defaultSettings;
+type BunyanPlugin = (
+  opts: Object,
+  pluginUtilities: PluginUtilities,
+) => { logger?: Logger };
+
+const bunyanPlugin: BunyanPlugin = (opts, { requireModule }) => {
```
threehams, line 36: I wonder if the future of plugins should include some first-class configuration stuff.

threehams (approved):

PR #1169, opened 2018-03-28
Add Flow library definitions for Webpack-aliased imports.

threehams:
Gluestick uses lots of Webpack aliases to reference files in the user project. The Flow error (unknown module) was suppressed for all of them.

This takes all the existing typings and moves them to library definitions, so that imports are correctly typed.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -10,8 +10,8 @@
 [include]
 
 [libs]
-flow/.*
-flow-typed/.*
+flow/
+flow-typed/
 
 [options]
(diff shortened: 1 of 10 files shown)


.flowconfig
```diff
@@ -10,8 +10,8 @@
 [include]
 
 [libs]
-flow/.*
-flow-typed/.*
+flow/
```
threehams, line 13: `.*` causes library definitions to fail in all kinds of interesting ways.

threehams:
[clabot:check]

PR #1171, opened 2018-03-29
Move static imports to middleware.js and set up test mocks.

threehams:
Refactor main.js, middleware.js, and renderer.js to import aliased files instead of passing them along from one to the other. The goal is to have middleware match the signature of Express middleware of `(request, response)`. This is easier to understand, but also sets us up to use webpack-hot-server-middleware.

Stop renaming arguments as they flow through each file. Make import naming more consistent.
Update tests to allow mocking of alias imports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "moduleNameMapper": {
       "^application-config$": "<rootDir>/packages/gluestick/src/__tests__/mocks/applicationConfig.js",
       "^entry-wrapper$": "<rootDir>/packages/gluestick/src/__tests__/mocks/EntryWrapper.js",
+      "^gluestick-hooks$": "<rootDir>/packages/gluestick/src/__tests__/mocks/gluestickHooks.js",
       "^redux-middlewares$": "<rootDir>/packages/gluestick/src/__tests__/mocks/reduxMiddlewares.js",
       "^project-entries$": "<rootDir>/packages/gluestick/src/__tests__/mocks/projectEntries.js",
       "^project-entries-config$": "<rootDir>/packages/gluestick/src/__tests__/mocks/projectEntriesConfig.js"
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "install:yarn:ci": "yarn install --frozen-lockfile && yarn run build && yarn run yarn:install:ci && lerna bootstrap --concurrency=1",
     "clean": "lerna clean --yes",
     "test": "jest --bail --no-cache -i",
+    "test:watch": "jest --watch",
     "test-coverage": "npm run test -- --coverage",
     "lint": "eslint ./packages",
     "flow": "flow",
@@ -64,7 +65,10 @@
     "transform": {
       "\\.js$": "<rootDir>/jestTransform.js"
     },
-    "testRegex": "(src|shared)/(.*/)?__tests__/.*\\.test\\.js$"
+    "testRegex": "(src|shared)/(.*/)?__tests__/.*\\.test\\.js$",
+    "moduleNameMapper": {
+      "project-entries": "<rootDir>/packages/gluestick/src/__tests__/mocks/context.js"
+    }
   },
   "private": true,
   "workspaces": [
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -64,7 +65,15 @@
     "transform": {
       "\\.js$": "<rootDir>/jestTransform.js"
     },
-    "testRegex": "(src|shared)/(.*/)?__tests__/.*\\.test\\.js$"
+    "testRegex": "(src|shared)/(.*/)?__tests__/.*\\.test\\.js$",
+    "moduleNameMapper": {
```
threehams, line 69: I can't leave comments in JSON, but these act as auto-mocks. This appears to be the only way to mock something which aliases to a nonexistent file (these only exist once they're in a project).

packages/gluestick/src/__tests__/mocks/EntryWrapper.js
```diff
@@ -0,0 +1 @@
+module.exports = () => {};
```
threehams, line 1: These are the minimum to get the tests running. (Likely they'll need to be expanded over time.)

packages/gluestick/src/renderer/__tests__/middleware.test.js
```diff
@@ -73,84 +69,56 @@ const response: Response = {
...
   preRenderFromCache: jest.fn(v => v),
   postRenderRequirements: jest.fn(v => v),
   preRedirect: jest.fn(v => v),
   postRenderProps: jest.fn(v => v),
   postGetCurrentRoute: jest.fn(v => v),
   postRender: jest.fn(v => v),
   error: jest.fn(v => v),
-};
-
-const options = {
```
threehams, line 86: All these consts are now covered by the mocks.

packages/gluestick/src/renderer/main.js
```diff
@@ -136,24 +120,11 @@ module.exports = function startRenderer({ config, logger }: Context) {
...
-            reduxMiddlewares,
-            thunkMiddleware,
-            reduxEnhancers,
-            envVariables,
-            httpClient: applicationConfig.httpClient || {},
-            entryWrapperConfig: {},
-          },
-          { hooks, hooksHelper: hooksHelper.call },
+        return middleware({ config, logger }, req, res, {
+          assets,
```
threehams, line 124: assets: comes from a cached file. `webpack-flush-chunks` changes this.
hooks: requires serverPlugins to create, which is completely static _except_ for `logger`. Will be imported directly after `logger` becomes a singleton.
serverPlugins: same, reliant on `logger`.

packages/gluestick/src/renderer/main.js
```diff
@@ -136,24 +120,11 @@ module.exports = function startRenderer({ config, logger }: Context) {
 
     readAssets(assetsFilename)
       .then((assets: Object): Promise<void> => {
-        return middleware(
-          { config, logger },
-          req,
-          res,
-          { entries, entriesConfig, entriesPlugins: runtimePlugins },
```
threehams, line 143: note the crazy renaming here! entriesPlugins is really runtimePlugins. hooksHelper is really hooksHelper.call. Knowing this will help the middleware.js changes make more sense.

packages/gluestick/src/renderer/middleware.js
```diff
@@ -24,75 +22,83 @@ const errorHandler = require('./helpers/errorHandler');
 const getCacheManager = require('./helpers/cacheManager');
 const getStatusCode = require('./response/getStatusCode');
 const createPluginUtils = require('../plugins/utils');
+const hooksHelper = require('./helpers/hooks');
+
+const entries = require('project-entries').default;
```
threehams, line 27: Many of these will be removed on the next ticket, as they're just passed along to `render.js`.

packages/gluestick/src/renderer/middleware.js
```diff
@@ -24,75 +22,83 @@ const errorHandler = require('./helpers/errorHandler');
...
+    hooks: GSHooks,
+    serverPlugins: ?(ServerPlugin[]),
+    assets: Object,
   },
-  { hooks, hooksHelper }: { hooks: GSHooks, hooksHelper: Function },
-  serverPlugins: ?(ServerPlugin[]),
-) {
+) => any;
+
+const middleware: Middleware = async (
```
threehams, line 55: Changed to arrow function for nicer typing. Renamed to match filename.

christinebrass (approved):

PR #1173, opened 2018-04-02
Change logger to be a singleton and move to imports rather than DI.

threehams:
`logger` was set up with a factory inside of `entry.js`, but didn't take advantage of the factory function. Instead, it just passed `logger` along to everything, resulting in `logger` showing up in the signatures of dozens of functions - despite having no external dependencies.

This PR moves the logger from `entry.js` to `logger.js`, and does the same with application config (also static). I've also simplified `prepareServerPlugins` to `serverPlugins` as an example of how this can be used.


packages/gluestick/src/renderer/main.js
```diff
@@ -1,19 +1,5 @@
 /* @flow */
-
-/**
- * To import/require file from project use aliases:
```
threehams: This comment didn't apply, even before these changes.

packages/gluestick/src/renderer/main.js
```diff
@@ -1,19 +1,5 @@
...
- *   root, src, actions, assets, components, containers, reducers, config
- * To import/require renderer server file use relative paths.
- */
-/* eslint-disable prefer-arrow-callback */
```
threehams, line 8: This was disabling all of eslint inside this file, and didn't apply anyway.

packages/gluestick/src/renderer/__tests__/entry.test.js
```diff
@@ -1,36 +0,0 @@
-jest.mock('../main', () => jest.fn());
```
threehams, line 1: This file no longer does any of this.

packages/gluestick/src/plugins/serverPlugins.js
```diff
@@ -51,7 +53,7 @@ const compilePlugin = (
  * Compile server plugins from given array. Those plugins are compiled with provided by user
  * `options` object from plugin declaration file, and utilities from gluestick like logger.
  */
-module.exports = (logger: BaseLogger, plugins: PluginRef[]): ServerPlugin[] => {
+const prepareServerPlugins = (): ServerPlugin[] => {
```
christinebrass, line 56: 👍

christinebrass (approved):
LGTM!

PR #1174, opened 2018-04-03
Fix bunyan logging options

christinebrass:
Fix use of the spread operator to get the correct user-configured Bunyan options.

### DIFF
--- a/packages/gluestick-plugin-bunyan/src/__tests__/server.test.js
+++ b/packages/gluestick-plugin-bunyan/src/__tests__/server.test.js
@@ -2,7 +2,6 @@
 
 import invariant from 'invariant';
 import gluestickPluginBunyan from '../server.js';
-import type { Logger } from '../server.js';
 
 describe('Bunyan plugin', () => {
   let mockLogger;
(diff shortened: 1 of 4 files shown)


packages/gluestick-plugin-bunyan/src/__tests__/server.test.js
```diff
@@ -52,15 +56,14 @@ describe('Bunyan plugin', () => {
         defaultLogger: mockLogger,
       },
     );
-    // Force a failure for Flow: https://github.com/facebook/jest/issues/3764
-    if (!logger) {
-      expect(false).toBeTruthy();
-      return;
-    }
+    // https://github.com/facebook/jest/issues/3764
+    invariant(logger, 'logger is not optional in this test');
```
threehams: Flow and `invariant` go really well together, so this can be really concise. Can take out the manual typing, too.
```
    const { logger } = gluestickPluginBunyan(
      {},
      {
        requireModule: mockRequireModule,
        defaultLogger: mockLogger,
      },
    );
    invariant(
      logger && logger.fields && logger.fields.name,
      'logger.fields.name is required',
    );
    expect(logger.fields.name).toEqual('default name');
```

The `invariant` call removes the need for `expect(logger).toBeDefined();`. The message is readable enough when anything fails in the `invariant` call:
```
  ● Bunyan plugin › should return the default logger if the config is empty

    Invariant Violation: logger.fields.name is required

      at invariant (packages/gluestick-plugin-bunyan/node_modules/invariant/invariant.js:40:15)
      at Object.it (packages/gluestick-plugin-bunyan/src/__tests__/server.test.js:55:29)
```

christinebrass: Nice!

threehams (approved):

PR #1178, opened 2018-04-20
Remove gluestick-config-legacy.

threehams:
gluestick-config-legacy is for migration to 1.0.0, and was always meant to be temporary. Maintaining it is becoming difficult.
This is a breaking change for Gluestick 3.0.0.


toddw (approved):
:+1:

PR #1180, opened 2018-04-20
Refactor out universal-webpack.

threehams:
universal-webpack's externals function is incompatible with react-universal-component and our own aliases. This removes it in favor of a simpler approach that reads the filesystem and marks all node_modules as external.


packages/gluestick/src/config/webpack/webpack.config.server.js
```diff
@@ -71,5 +73,19 @@ module.exports = (
...
   );
-  return serverConfiguration(config, settings);
+  const universal = serverConfiguration(config, settings);
+
+  // overwrite universal-webpack's way of doing externals as it
+  // doesn't account for aliases, and breaks react-universal-component
+  universal.externals = fs
+    .readdirSync(path.join(process.cwd(), 'node_modules'))
+    .filter(
+      x => !/\.bin|react-universal-component|webpack-flush-chunks/.test(x),
```
threehams: This references files which don't exist yet, but they won't cause problems.

threehams:
[clabot:check]

christinebrass (approved):
LGTM

PR #1187, opened 2018-04-26
Write to both project and gluestick build folders with watch.

threehams:
Prevents a case where watch writes to the project build folder, but not to its own, and reinstalling node_modules installs partially/completely stale code.

This also includes refactoring to have better variable names than `e`, as it took a long time to figure out what was going on in here.

### DIFF
--- a/packages/gluestick-cli/src/watch.js
+++ b/packages/gluestick-cli/src/watch.js
@@ -4,12 +4,12 @@ const chokidar = require('chokidar');
 const fs = require('fs-extra');
 
 module.exports = exitWithError => {
-  const packagePath = path.join(process.cwd(), 'package.json');
+  const packageJsonPath = path.join(process.cwd(), 'package.json');
   let packageContent = null;
   try {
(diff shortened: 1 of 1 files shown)


packages/gluestick-cli/src/watch.js
```diff
@@ -67,65 +67,81 @@ module.exports = exitWithError => {
...
-    const packageName = gsPackages[i];
+  gsDependenciesPath.forEach((packagePath, index) => {
+    const packageName = gsPackages[index];
     const packageBabelRc = JSON.parse(
-      fs.readFileSync(path.join(e, '.babelrc')).toString(),
+      fs.readFileSync(path.join(packagePath, '.babelrc')).toString(),
     );
-    const convertFilePath = filePath => {
+    const fileList = filePath => {
+      if (/[\\/]build[\\/]/.test(filePath)) {
```
christinebrass: Skip build folders?

christinebrass (approved):

PR #1188, opened 2018-04-27
Add snapshot tests for Webpack configuration.

threehams:
Add snapshot tests for Webpack config. This does not include config options which produce side effects and will be removed.

Changed default test runner options. Existing options were only useful for CI, and had negative effects on local development (including continually updating snapshot tests).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
       - attach_workspace:
           at: ~/gluestick
 
-      - run: yarn run test -i --coverage
+      - run: yarn run test --runInBand --coverage
 
       - store_artifacts:
           path: coverage
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
       - attach_workspace:
           at: ~/gluestick
 
-      - run: yarn run test --coverage
+      - run: yarn run test -i --coverage
 
       - store_artifacts:
           path: coverage
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "mkdirp": "0.5.1",
     "node-fetch": "^1.7.1",
     "prettier": "^1.5.3",
-    "rimraf": "2.6.1"
+    "rimraf": "2.6.1",
+    "traverse": "0.6.6"
   },
   "jest": {
     "notify": true,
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       - attach_workspace:
           at: ~/gluestick
 
-      - run: yarn run test --coverage
+      - run: yarn run test -i --coverage
```
christinebrass: Generally I prefer the non-aliased version of flags when it exists in a file: `--runInBand`. I wonder why we previously had the `--no-cache` option turned on, and whether we were running into issues there? I guess we'll find out. Also why not bail right away on CI? I'd expect if one test failed, a dev would run the whole suite when fixing the problematic first failed test.

package.json
```diff
@@ -8,7 +8,7 @@
     "install:yarn": "yarn install && yarn run build && yarn run yarn:install && lerna bootstrap --concurrency=1",
     "install:yarn:ci": "yarn install --frozen-lockfile && yarn run build && yarn run yarn:install:ci && lerna bootstrap --concurrency=1",
     "clean": "lerna clean --yes",
-    "test": "jest --bail --no-cache -i",
+    "test": "jest",
```
christinebrass, line 11: 👍

.circleci/config.yml
threehams: Agreed, this was a copy/paste. Changing

packages/gluestick/src/config/__tests__/__snapshots__/compileWebpackConfig.test.js.snap
```diff
@@ -0,0 +1,937 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`config/compileWebpackConfig in dev environment returns the correct client config 1`] = `
+Object {
+  "context": "/Users/dedmondson/src/tc/gluestick",
```
christinebrass: I suspect these file paths are causing the CI issue.

threehams: They are. This may not be easy to fix. Mocking path.resolve breaks a lot.

PR #1190, opened 2018-05-02
Refactor out shared Webpack config.

threehams:
Remove the shared config, so that the starting point for Webpack config is "client" and "server" instead of "shared." This removes a lot of object override code.

This introduces duplication, some of which will be refactored out in the next PR (especially aliases). There will be some intentional duplication after all this work.


packages/gluestick/src/config/__tests__/__snapshots__/compileWebpackConfig.test.js.snap
```diff
@@ -153,7 +154,12 @@ Object {
                       "transform-flow-strip-types",
                     ],
                     "presets": Array [
-                      "es2015",
+                      Array [
```
threehams, line 148: Let me do a quick check in the app. This should be fine.

packages/gluestick/src/config/__tests__/__snapshots__/compileWebpackConfig.test.js.snap
```diff
@@ -232,6 +238,34 @@ Object {
               "test": /\\\\\\.\\(scss\\|css\\)\\$/,
             },
           },
+          NormalModuleReplacementPlugin {
```
threehams: ChunksPlugin is really weird and requires a Webpack config to be passed into it. I'm glad we're getting rid of this soon.

Looking through the universal-webpack code, supplying extra plugins shouldn't do anything.

christinebrass (approved):
👍

PR #1191, opened 2018-05-02
add support for async server plugin

toddw:


threehams:
We'll have to figure out how to make this work with react-universal-component, since it also relies on the caching/flushing technique.

threehams:
Have you noticed the tests are already set up to use async/await? It looks like we did something similar to this in the past.

threehams (approved):

PR #1192, opened 2018-05-02
add support for runtime plugin options

toddw:
Currently options are not passed to runtime plugins when defined in _gluestick.plugins.js_, this changes that so that they will be.

### Special Note
Per requests on the PR this now also refactors the shape of a Plugin to have a more unified flow type. This makes this PR much more risky than it was and therefore will require some extra manual testing before it should be merged
- [x] manually testing complete


packages/gluestick/src/generator/templates/EntryWrapper.js
```diff
@@ -143,7 +143,9 @@ export default class EntryWrapper extends Component {
     } = this.props;
 
     return rootWrappers.reduce((prev, curr) => {
-      return curr(prev, rootWrappersOptions);
+      const plugin = curr.ref.plugin || curr;
```
threehams: 👍 I'm not crazy about `ref` since I don't think it's common enough to be self-documenting, but you're keeping consistency with other plugins which is much more important.

packages/gluestick/src/renderer/render.js
```diff
@@ -60,9 +60,7 @@ module.exports = function render(
   );
   const isEmail = !!currentRoute.email;
   const routerContext = <RouterContext {...renderProps} />;
-  const rootWrappers = entriesPlugins
-    .filter(plugin => plugin.meta.wrapper)
-    .map(({ plugin }) => plugin);
+  const rootWrappers = entriesPlugins.filter(plugin => plugin.ref.meta.wrapper);
```
threehams: I think I'm finally starting to understand some of these plugin lines after this PR.

packages/gluestick/src/renderer/render.js
```diff
@@ -29,7 +29,7 @@ type WrappersRequirements = {
   BodyWrapper: Object,
   entryWrapperConfig: Object,
   envVariables: any[],
-  entriesPlugins: { plugin: Function, meta: Object }[],
+  entriesPlugins: { ref: { plugin: Function, meta: Object }, options?: Object }[],
```
threehams: Do we have better typing for `meta` or is it open-ended?

threehams (commented):
A good PR description would have helped a lot here. This isn't expanding plugins with more options, but rather bringing runtime plugins in line with server plugins, which use the same structure.

packages/gluestick/src/renderer/render.js
```diff
@@ -29,7 +29,7 @@ type WrappersRequirements = {
   BodyWrapper: Object,
   entryWrapperConfig: Object,
   envVariables: any[],
-  entriesPlugins: { plugin: Function, meta: Object }[],
```
threehams, line 32: I think you can improve this, and it'll help the plugin system a lot.

Flow makes Gluestick less readable, but it's not because of Flow. It's because of the large number of inlined types which get repeatedly redefined, as well as the types that are just _slightly_ different from existing types. In many cases, there's no reason for the slight differences.

In `src/types.js`, these are types for plugins. (I'm not sure what `Plugin` means in this context vs the others.)
```
export type Plugin = {
  name: string,
  meta: {
    [key: string]: any,
  },
  body: Function | null,
  options: Object,
};

export type ConfigPlugin = {
  name: string,
  meta: {
    [key: string]: any,
  },
  preOverwrites: {
    sharedWebpackConfig?: (config: WebpackConfig) => WebpackConfig,
  },
  postOverwrites: {
    gluestickConfig?: (config: GSConfig) => void,
    clientWebpackConfig?: (config: WebpackConfig) => WebpackConfig,
    serverWebpackConfig?: (config: WebpackConfig) => WebpackConfig,
    vendorDllWebpackConfig?: (config: WebpackConfig) => WebpackConfig,
  },
};

export type RuntimePlugin = {
  name: string,
  meta: {
    [key: string]: any,
  },
  body: {
    rootWrapper?: (component: Object) => Object,
  },
};

export type ServerPlugin = {
  name: string,
  meta: {
    [key: string]: any,
  },
  renderMethod: Function,
  hooks: GSHooks,
  logger: Logger,
};
```

Then there's a PluginRef from serverPlugins.js. I don't understand how or why the above type gets converted to this structure.
```
type PluginRef = {
  ref: Function,
  type: string,
  options?: Object,
};
```

and finally your inlined type, which seems to be different from all the above:
```
{ ref: { plugin: Function, meta: Object }, options?: Object }
```

I'm not opposed to having different types and signatures for different types of plugins (it prevents large numbers of optional types) as long as we keep these types intact as long as possible. Can you bring some sanity to these, and see if you can figure out how to avoid the addition of a new type specifically for `render.js`?

threehams (changes requested):

packages/gluestick/src/renderer/render.js
toddw: I totally agree with this. However, it sounds like something that should be done in a separate pull request. If this PR introduced the issue then it would make sense to fix it here but it just makes due with that current landscape. However, there may be an opportunity to clean up the types that were touched without adding too much scope to this PR. Let me see what I can do/

packages/gluestick/src/generator/predefined/serverEntries.js
```diff
@@ -31,15 +31,21 @@ ${args =>
 
 export const plugins = [
   ${args =>
-    args.plugins.reduce((prev, curr, index, arr) => {
+    args.plugins.reduce((prev, curr, index) => {
```
threehams, line 34: Today I learned reduce() can accept a fourth argument!

threehams (approved):

PR #1194, opened 2018-05-15
Set up ability to inline CSS into server response.

threehams:
Save a round trip for CSS. Makes a big difference when initial load times are more important than return visits.

### DIFF
--- a/packages/gluestick/src/config/webpack/ChunksPlugin.js
+++ b/packages/gluestick/src/config/webpack/ChunksPlugin.js
@@ -1,29 +1,20 @@
 /* @flow */
-import type { WebpackConfig } from '../../types';
+import type { ChunksInfo, ChunkInfo, WebpackConfig } from '../../types';
 
 const path = require('path');
 const fs = require('fs');
 const mkdir = require('mkdirp');
 
-type ChunksInfo = {
-  javascript: {
-    [key: ?string]: string,
-  },
-  styles: {
-    [key: ?string]: string,
-  },
-};
-
 const chunkInfoFilePath = (
   webpackConfiguration: WebpackConfig,
   chunkInfoFilename?: string = 'webpack-chunks.json',
 ): string => {
-  // $FlowIgnore `output` is a Object
   return path.join(webpackConfiguration.output.path, chunkInfoFilename);
 };
 
-const getChunksInfoBody = (json: Object, publicPath: string): ChunksInfo => {
-  const assetsByChunk: Object = json.assetsByChunkName;
+const getChunksInfoBody = (stats: Object, publicPath: string): ChunksInfo => {
+  const assetsByChunk = stats.assetsByChunkName;
+  console.log(assetsByChunk);
 
   const assetsChunks: ChunksInfo = {
     javascript: {},
@@ -31,8 +22,8 @@ const getChunksInfoBody = (json: Object, publicPath: string): ChunksInfo => {
   };
 
   // gets asset paths by name and extension of their chunk
-  const getAssets = (chunkName: string, extension: string): string[] => {
-    let chunk: string | string[] = json.assetsByChunkName[chunkName];
+  const getAssets = (chunkName: string, extension: string): ChunkInfo[] => {
+    let chunk: string | string[] = stats.assetsByChunkName[chunkName];
 
     // a chunk could be a string or an array, so make sure it is an array
     if (!Array.isArray(chunk)) {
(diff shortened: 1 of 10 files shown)


packages/gluestick/src/renderer/helpers/linkAssets.js
```diff
@@ -58,24 +75,43 @@ module.exports = function linkAssets(
...
   if (stylesHref) {
-    styleTags.push(
-      <link key={key++} rel="stylesheet" type="text/css" href={stylesHref} />,
-    );
+    if (config.GSConfig.inlineAllCss) {
+      const contents = await memoizedRead(
+        path.join(process.cwd(), 'build', stylesHref),
+      );
+      styleTags.push(
+        <style dangerouslySetInnerHTML={{ __html: contents.toString() }} />,
```
sethbattin: I might be mistaken, but I think recent `react-helmet` allows `style`s with text as children.

threehams: It does, but we're really low-level here (initial server render), and Helmet isn't really designed for order-dependent stuff like styles.

packages/gluestick/src/config/webpack/ChunksPlugin.js
```diff
@@ -41,18 +32,18 @@ const getChunksInfoBody = (json: Object, publicPath: string): ChunksInfo => {
...
     return chunk
       .filter(name => path.extname(name) === `.${extension}`)
-      .map(name => `${publicPath}${name}`);
+      .map(name => ({ url: `${publicPath}${name}`, name }));
   };
 
   Object.keys(assetsByChunk).forEach((name: string) => {
     // The second asset is usually a source map
-    const jsAsset: string = getAssets(name, 'js')[0];
+    const jsAsset = getAssets(name, 'js')[0];
```
hhtran, line 40: why no more flow typing?

packages/gluestick/src/config/webpack/ChunksPlugin.js
```diff
@@ -22,17 +13,17 @@ const chunkInfoFilePath = (
   return path.join(webpackConfiguration.output.path, chunkInfoFilename);
 };
 
-const getChunksInfoBody = (json: Object, publicPath: string): ChunksInfo => {
-  const assetsByChunk: Object = json.assetsByChunkName;
+const getChunksInfoBody = (stats: Object, publicPath: string): ChunksInfo => {
```
threehams, line 16: ChunksPlugin is from universal-webpack and was copy/pasted in here a while ago. It tracks JS + CSS chunks and writes the information to a file. I modified it to include the original filename in addition to the URL.

packages/gluestick/src/renderer/helpers/linkAssets.js
```diff
@@ -1,20 +1,25 @@
 /* @flow */
 
-import type { Context } from '../../types';
+import type { ChunksInfo, Context } from '../../types';
 
 const React = require('react');
 const path = require('path');
 const fs = require('fs');
+// $FlowIgnore promisify is not available in this version of Flow
```
threehams, line 8: Upgrading Flow will flag this as an error. We can remove it at that time.

packages/gluestick/src/renderer/helpers/linkAssets.js
```diff
@@ -26,69 +31,85 @@ const filterEntryName = (name: string): string => {
...
+
+const memoizedRead = async (name: string) => {
+  if (cache[name]) {
+    return cache[name];
+  }
+
+  const localPath = path.join(process.cwd(), 'build', 'assets', name);
+  const contents = await readFileAsync(localPath);
+
+  cache[name] = contents.toString();
```
threehams, line 54: This will definitely increase server memory usage, but will not cause garbage collection calls. I'll monitor it after release.

packages/gluestick/src/renderer/helpers/linkAssets.js
```diff
@@ -26,69 +31,85 @@ const filterEntryName = (name: string): string => {
...
   entryPoint: string,
   assets: Object,
   loadjsConfig: Object,
-): { styleTags: Object[], scriptTags: Object[] } {
-  const styleTags: Object[] = [];
-  const scriptTags: Object[] = [];
-  let key: number = 0;
-  const entryPointName: string = filterEntryName(entryPoint);
+) {
+  const styleTags = [];
```
threehams, line 64: I removed a lot of unnecessary Flow typing from this file. All of this is inferred.

packages/gluestick/src/renderer/helpers/linkAssets.js
jacobmoretti: Curious how you''ll monitor this. I haven't looked into it too much on NR but do we get pretty good visibility into the health of our node instances through NR?

packages/gluestick/src/config/webpack/ChunksPlugin.js
threehams: commented a bit late: most of the Flow types in this file are redundant and already inferred. I improved typing overall by getting rid of some `Object` ones.

packages/gluestick/src/renderer/helpers/linkAssets.js
christinebrass: I think we can leverage datadog for this.

christinebrass (approved):

packages/gluestick/src/renderer/helpers/linkAssets.js
threehams: Today I learned that datadog exists.

PR #1201, opened 2018-06-26
Allow custom status code through location.state.

threehams:
This allows 302 redirects (or other status codes) if `location.state.status` exists.


christinebrass (approved):

packages/gluestick/src/renderer/middleware.js
```diff
@@ -124,8 +124,10 @@ const middleware: Middleware = async (req, res, { assets }) => {
     );
     if (redirectLocation) {
       callHook(hooks.preRedirect, redirectLocation);
+      const status =
+        (redirectLocation.state && redirectLocation.state.status) || 301;
```
threehams, line 128: `state` defaults to null or undefined. I wish it didn't.

PR #1204, opened 2018-07-11
Lock down autoprefixer to 6.7.7

christinebrass:
We should lock down dependencies to reduce risk.


hhtran (approved):
Unrelated, but should we lock down other dependencies at some point? We have a lot that aren't locked down

christinebrass:
I think we should. We'll have to evaluate what versions are actually in use and lock to those.

PR #1205, opened 2018-07-11
Update yarn.lock

christinebrass:


hhtran (approved):
My approval has no weight, but I still give my 👍 of approval haha

PR #1218, opened 2018-12-04
Add custom cache key strategy support

hhtran:
Sometimes apps need certain routes to be able to have custom cache key support (e.g. to be able to separately cache certain route query params as unique cached payloads). As of the current gluestick, only the hostname and pathname are used in determining unique cache keys.

This PR adds in the ability for users to define a `cacheKey` prop on their routes, which is a function that takes in a request object and returns a cache key.

This is intended to be a non-breaking change, with the current caching behavior remaining intact, with only routes containing the new `cacheKey` prop now seeing different behavior

### DIFF
--- a/packages/gluestick/src/__tests__/mocks/context.js
+++ b/packages/gluestick/src/__tests__/mocks/context.js
@@ -96,6 +96,7 @@ const request: Request = {
   url: '/',
   headers: { 'user-agent': '' },
   method: 'GET',
+  query: {},
 };
 
 const entriesConfig: EntriesConfig = {
--- a/packages/gluestick/src/renderer/__tests__/render.test.js
+++ b/packages/gluestick/src/renderer/__tests__/render.test.js
@@ -82,6 +82,7 @@ describe('renderer/render', () => {
     url: '',
     hostname: '',
     method: 'GET',
+    query: {},
   };
 
   const store = {
--- a/packages/gluestick/src/renderer/helpers/__tests__/cacheManager.test.js
+++ b/packages/gluestick/src/renderer/helpers/__tests__/cacheManager.test.js
@@ -46,9 +46,19 @@ describe('renderer/helpers/cacheManager', () => {
       true,
     );
     const cache = {
-      get: jest.fn(key => (key === 'localhost/' ? 'output' : null)),
+      get: jest.fn(key => {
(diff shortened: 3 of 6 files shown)


hhtran:
Confirmed locally that one can set the cacheKey prop on a Route and get the keys seperately cached:

```
<Route  cacheKey={
    req => {
        return `${req.hostname}-${req.url}-${req.query.viprStyleId}`;
    }
} {...rest} />

```

```
DEBUG: Set cache: www.local.truecardev.com/prices-new/acura/ilx-pricing/?viprStyleId=60564
DEBUG: Set cache: www.local.truecardev.com/prices-new/acura/ilx-pricing/?viprStyleId=60563
DEBUG: Get cached: www.local.truecardev.com/prices-new/acura/ilx-pricing/?viprStyleId=60564
```

threehams:
Since we're starting to use this more, I'd like there to be a configurable way to bypass the cache. The most critical example (for us) is if `state.currentUser.loaded` is true, but we should be able to configure it for any state slice potentially containing PII.

hhtran:
Confirmed that you can have the cache not set when the current user is loaded:
```
                cacheKey={(req, state) => {
                  if (state && state.currentUser && state.currentUser.loaded) {
                    return null;
                  }
                  return `${req.hostname}-${req.url}-${req.query.viprStyleId}`;
                }}
```

Verified locally that other things are cached as expected

hhtran:
Also, I'm not so sure that we should have a separate `shouldCache` prop. It would be very similarly named to have both a `cache={true} and a shouldCache={SOME_FUNC_THAT_MIGHT_EVAL_AS_FALSE}`. 

And I can't think of any use cases where you would want to return `null` for your cache key and still want your stuff cached. Using the cacheKey func as the main determinant for caching/not caching things would also allow the usage of the cache to fail-safe in the case that the user has a falsy return value from the cacheKey function. If the user forgets to return anything, we cache nothing instead of mapping everything to the same payload

```
var x = {};
const cacheKey = () => {};
x[cacheKey()] = "hello";
console.log(x); // 11:28:54.455 {undefined: "hello"}
```

threehams (approved):
Yep. Do it.

There's some cleanup possible here, but it makes more sense to do it in the app. Specifically the six six unnamed arguments accepted by `setCache`.

PR #1221, opened 2019-01-15
Use req.hostname for hostname query in axios client

christinebrass:
This will ensure that the value of X-Forwarded-Host is used whenever `trust proxy` is enabled, falling back to the normal host header when it is disabled.

### DIFF
--- a/packages/gluestick/shared/lib/getHttpClient.js
+++ b/packages/gluestick/shared/lib/getHttpClient.js
@@ -47,7 +47,7 @@ export default function getHttpClient(
   // If a request object is provided, then we want to merge the custom headers
   // with the headers that we sent from the browser in the request.
   client = httpClient.create({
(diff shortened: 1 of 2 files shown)


packages/gluestick/shared/lib/getHttpClient.js
```diff
@@ -47,7 +47,7 @@ export default function getHttpClient(
   // If a request object is provided, then we want to merge the custom headers
   // with the headers that we sent from the browser in the request.
   client = httpClient.create({
-    baseURL: protocol + req.headers.host,
+    baseURL: protocol + req.hostname,
```
christinebrass, line 50: This will fall back to `req.get('host')` when `trust proxy` is not enabled: 

https://github.com/expressjs/express/blob/4.16.4/lib/request.js#L431-L433

threehams (approved):
Matches app changes.

PR #1222, opened 2019-01-16
Patch express req.host to include port number

threehams:
Express has a long-standing bug in v4 where the port number is dropped from `host`, which is why `req.host` is marked as deprecated. In v5, `req.host` is fixed and `req.hostname` does not include the port.

For now, we're using middleware to patch `req.host` to include the port number, which is correct behavior.


threehams:
Approving on behalf of @christinebrass who wrote the code (I'm just opening the PR).

christinebrass (approved):

